PR #354, opened 2014-12-02
Open Humans integration and data access work

beaugunderson:
Just creating a WIP pull request for Open Humans integration work that will happen this week.

- [x] add quick-start instructions to `README.md`
- [x] comment out and note that insertAGMultiple is vulnerable to SQL injection (it's currently unused)
- [x] consolidate all data access into `amgut.connections`
- [x] lazily connect to databases (obviates need to disconnect the automatically-connected connections in `scripts/ag`, for example)
- [x] increase robustness of `scripts/ag`
- [x] make sure everything is created in the `ag` schema (including the stored procedures and the `settings` table)
- [x] always specify SEARCH_PATH as "ag, public"
- [x] add a Procfile for ease of running with a specified port
- [x] add labels for login/password field on the main page
- [x] resize the password field to fit a bcrypt hash
- [x] store and verify passwords as bcrypt hashes
- [x] add the Open Humans access token
- [x] investigate an alternate approach for lazy-loading database connections (use `__getattr__` to connect on first attribute access)
- [x] add a command to `scripts/ag` to migrate unencrypted passwords
- [x] add a command or database patch to migrate settings to the `ag` schema
- [x] add a command or database patch to migrate stored procedures to the `ag` schema

### Migration steps

- `scripts/ag patch` (this is idempotent)
- `scripts/ag migrate_passwords` (this is idempotent)


amgut/connections.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# -----------------------------------------------------------------------------
+#
+# This module handles lazy-loading the various connections needed. This
+# simplifies code that relies on amgut (like scripts/ag) but that does not need
+# connections to the database since it handles them itself.
+
+
+class _LazyAgData(object):
```
wasade: This is a very clean way to handle these connections. @josenavas @squirrelo, just pinging the two of you as it might be nice to adopt something like this in qiita

amgut/db/populate_test.sql
```diff
@@ -1,6 +1,83 @@
...
+        ('d8592c74-7da3-2135-e040-8a80115d6401','d8592c74-7da2-2135-e040-8a80115d6401','000000001','000000001.jpg','','Stool',
+            '12/02/2013','foo','06:20 AM','','','N','N','N','','None','Y','N','N'),
+        ('ded369ec-7d0f-4d8c-e040-8a80115d6742','d8592c74-7da2-2135-e040-8a80115d6401','000006616','000006616.jpg','','Nasal mucus','08/12/2014',
+            'MAKE SURE THIS IS BEING PASSED','12:05 PM','','','','','','','','','','');
+insert into ag_handout_kits (kit_id, password, barcode, verification_code, swabs_per_kit, print_results)
+    values('test_ha', '1234', '000000002', '5678', '3', 'n'),
+          ('test_ha', '1234', '000000003', '5678', '3', 'n'),
+          ('test_ha', '1234', '000000004', '5678', '3', 'n');
+insert into ag_participant_exceptions (ag_login_id, participant_name)
+    values ('d8592c74-7da1-2135-e040-8a80115d6401', 'execption');
```
wasade: minor, but typo: execption -> exception

amgut/lib/data_access/ag_data_access.py
```diff
@@ -189,6 +197,7 @@ def getAGLogins(self):
 
         return return_res
 
+    # note: only used by the password migration
```
wasade: ?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -428,14 +441,16 @@ def getConsent(self, survey_id):
...
 
-    def insertAGMultiple(self, ag_login_id, participant_name, field_name,
-                         field_value):
-        sql = ("insert into ag_survey_multiples (ag_login_id, "
-               "participant_name,item_name, item_value) values ('{0}','{1}',"
-               " '{2}', '{3}')").format(ag_login_id, participant_name,
-                                        field_name, field_value)
-        self.get_cursor().execute(sql)
-        self.connection.commit()
+    # NOTE: This method is vulnerable to SQL injection as-written but is not
```
wasade: Much of the data access code is carryover from the old mod_python + Oracle deployment, and there are blocks of code that slipped through that should be removed. If this is not in use then please delete it.

amgut/lib/util.py
```diff
@@ -169,8 +170,9 @@ def survey_vioscreen(survey_id):
     """Return a formatted text block and URL for the external survey"""
     tl = text_locale['human_survey_completed.html']
     embedded_text = tl['SURVEY_VIOSCREEN']
+    user_info = ag_data.get_person_info(survey_id)
     url = ("https://vioscreen.com/remotelogin.aspx?Key=%s&RegCode=KLUCB" %
-           url_escape(encrypt_key(survey_id)))
+           url_escape(encrypt_key(survey_id, user_info)))
```
wasade: @beaugunderson, can you comment as to why this was done? @squirrelo, pinging you to keep you in the loop since you're most familiar with this interaction.

wasade: ...should have simply scrolled a little further. Makes sense now, thanks. @squirrelo, it would still be useful if you could once over these particular changes though

amgut/lib/data_access/ag_data_access.py
wasade: I see now, ignore the comment

scripts/ag
```diff
@@ -105,32 +147,118 @@ def clean_test():
...
+                    'CASCADE')
+
+        click.echo('Revoking all privileges on all tables in ag')
+        cur.execute('REVOKE ALL PRIVILEGES ON ALL TABLES IN SCHEMA ag '
+                    'FROM ag_wwwuser CASCADE')
+
+        click.echo('Revoking all privileges on all functions in ag')
+        cur.execute('REVOKE ALL PRIVILEGES ON ALL FUNCTIONS IN SCHEMA ag '
+                    'FROM ag_wwwuser CASCADE')
+    except:
```
wasade: If one of the `execute`s excepts then the rest won't execute. If the intent is to ensure all these are executed regardless of if anything excepts (which seems to be the case given the `pass`) then this structure may need to be revisited. However, this does not hamper production and only test environments, so its exposure is limited.

scripts/ag
```diff
@@ -105,32 +147,118 @@ def clean_test():
...
+    click.echo('Dropping schema ag...')
+    cur.execute('DROP SCHEMA IF EXISTS ag CASCADE')
+
+    cur.close()
+    conn.close()
+
+    click.echo('Dropping database {}'.format(AMGUT_CONFIG.database))
 
-@env.command()
+    command = ['dropdb', '--if-exists']
```
wasade: Why fork instead of issuing via psycopg2? Just generally curious

wasade:
Done with static analysis. This looks very good, thank you for the contributions, only a few minor comments. I'll pull this down either over the weekend or first thing Monday to test locally. 

@adamrp @ElDeveloper @squirrelo @josenavas, is anyone else able to pull down an run locally as well? I'd like to get this merged into master by mid next week at the latest, as well as any other pending PRs that need to get in.

amgut/lib/data_access/ag_data_access.py
beaugunderson: The one place where it was in use is inside the animal survey, but the bulk of that survey is commented out (including the line that references this function); I figured it may be used again at some point from there but wanted to note that potential for SQL injection so that it could be fixed before it was used again. Should it just be removed?

amgut/lib/util.py
squirrelo: Not sure why the change to getting user info here, but works for me. It doesn't look like it changes any fundamental code.

scripts/ag
beaugunderson: This was done early on as I was still learning how the automatic connection code worked; I thought a connection was being held open (and that was preventing the database from being dropped). After I refactored to use the lazy-loaded connections that issue went away, but I think this did fix another issue in my system:

I setup my database so that the amgut user only had access to the amgut database. To delete the database via psycopg2 I don't think you can specify the database to connect to (since you'll then be prevented from dropping it). There are no other databases my user had access to from which to delete the database, but deleting with dropdb did work.

beaugunderson:
@wasade OK, I think I addressed all your comments :)

amgut/lib/util.py
beaugunderson: Just consolidating things that access the database in one place, was purely organizational.l :)

squirrelo: Cool :+1:

scripts/ag
beaugunderson: True, probably better to make these into functions and loop through them, `pass`ing on each--but seems OK to leave for now.

amgut/lib/data_access/ag_data_access.py
wasade: Yes please, thanks!

Flagging @adamrp just to make sure he saw this note re: animal survey. We
haven't redone it yet, so it's still carrying some fluff.

scripts/ag
wasade: Ah, good point, that makes sense. Thanks!

wasade:
@beaugunderson, just pulled down to test local. First thing that popped up:

```bash
./ag drop_test
/Users/mcdonadt/rs/american-gut-web/amgut/lib/config_manager.py:33: UserWarning: Extra main section option(s) found: 'debug'
  warnings.warn("Extra %s found: %r" % (set_type, extra))
Dropping user ag_wwwuser
Reassigning owned objects
Revoking all privileges on ag2
Revoking all privileges in schema public
Revoking all privileges on all tables in public
Revoking all privileges on all functions in public
Revoking all privileges in schema ag
Revoking all privileges on all tables in ag
Revoking all privileges on all functions in ag
Dropping user ag_wwwuser...
Dropping schema ag...
Dropping database ag2
dropdb: unrecognized option `--if-exists'
Try "dropdb --help" for more information.
$ dropdb --version
dropdb (PostgreSQL) 9.1.9
```

beaugunderson:
@wasade Ah, looks like `--if-exists` was added in postgresql 9.2; does american-gut-web target 9.1? (Worth updating the README if so, which I can do)

wasade:
I reset one of the `kit_password`s in the test database, and the performed the password migration. I was unable to login using that account after, indicating an invalid password. Below are the steps performed. I tried to login with the plain text which threw a 500 error (that's okay I think for now as this should never happen in practice). I also tried logging in following the migration and was met with an invalid kit ID or password.

```bash
$ echo "update ag.ag_kit set kit_password='foo' where supplied_kit_id='test';" | psql -d ag2
Expanded display is used automatically.

UPDATE 1
Time: 5.102 ms
$ echo "select kit_password from ag.ag_kit where supplied_kit_id='test';" | psql -d ag2
Expanded display is used automatically.

+--------------+
| kit_password |
+--------------+
| foo          |
+--------------+
(1 row)

Time: 2.452 ms
$ ./ag migrate_passwords
Connecting to postgres for amgut.connections.ag_data
kit 1111 is already a bcrypt hash
migrating kit test to a bcrypt hash
$ echo "select kit_password from ag.ag_kit where supplied_kit_id='test';" | psql -d ag2
Expanded display is used automatically.

+--------------------------------------------------------------+
|                         kit_password                         |
+--------------------------------------------------------------+
| $2a$12$IRfEN4AAkOjv7ck9ErGlYe5mqQphHwLx.lvisIrIStjEVFcSTmxQS |
+--------------------------------------------------------------+
(1 row)

Time: 2.310 ms
```

wasade:
Those were the only issues I encountered! re: version, that's a good question. On production, we're actually using 9.3.5, so my env is actually a little behind. I think it is okay to set the minimum to 9.2 (eg >= 9.2). Sound good?

beaugunderson:
@wasade Sounds good re: 9.2; and I'll investigate the password reset issue!

wasade:
Great, thanks! I'll be in the air shortly, but will re-review around 11:30
PT if you're able to get commits in before then.

beaugunderson:
@wasade OK, the password issue is fixed. I implemented migration before I added encryption to the rest of the data access functions; once I did the migration script was actually hashing the passwords twice. :)

wasade:
Great, thanks!

PR #360, opened 2014-12-08
Proper deletion of user surveys and rename of function

squirrelo:
This nulls out the survey ID before deleting a survey so a FK issue no longer occurs. Also renamed the function to say it deletes survey, not user itself.


antgonza:
@squirrelo could you solve conflicts?

squirrelo:
Ready

amgut/lib/data_access/ag_data_access.py
```diff
@@ -412,6 +412,11 @@ def deleteAGParticipant(self, ag_login_id, participant_name):
                    "survey_id = %s")
             curr.execute(sql, [survey_id])
 
+            # Reset survey attached to barcode(s)
+            sql = ("UPDATE ag_kit_barcodes SET survey_id = NULL WHERE "
```
antgonza: I know the tests are passing but I'm not sure if this will work in the live site, this based as in the DB survey_id is a not nullable FK, right? Perhaps I'm looking at this wrongly.

squirrelo: The tests would fail as well then, it's all the same database setup.

wasade: `survey_id` is nullable in `ag_kit_barcodes` as unassigned barcodes are not associated with a valid survey

antgonza: Thanks!

PR #373, opened 2015-02-23
Change to redirect to sitebase after registration

squirrelo:
We are seeing a user error pop up every once and a while and this may be why. Technically never log the user in during registration, so can't redirect to the authed portal. Instead, this changes it to redirect to the standard portal.

### DIFF
--- a/amgut/handlers/auth_handlers.py
+++ b/amgut/handlers/auth_handlers.py
@@ -42,7 +42,7 @@ class AuthRegisterHandoutHandler(BaseHandler):
             ag_login_id, skid, kitinfo['password'],
             kitinfo['swabs_per_kit'], kitinfo['verification_code'],
             printresults)
-        if success == -1:
+        if success != 1:
             self.redirect(media_locale['SITEBASE'] + '/db_error/?err=regkit')
             return
 
@@ -53,7 +53,7 @@ class AuthRegisterHandoutHandler(BaseHandler):
         for row in results:
             barcode = row[0]
             success = ag_data.addAGBarcode(ag_kit_id, barcode)
-            if success == -1:
+            if success != 1:
                 self.redirect((media_locale['SITEBASE'] +
                                '/db_error/?err=regbarcode'))
                 return
@@ -76,7 +76,8 @@ class AuthRegisterHandoutHandler(BaseHandler):
 
             self.render('help_request.html', skid=skid, result=result)
 
-        self.redirect(media_locale['SITEBASE'] + '/authed/portal/')
+        # added person but not logged in so just go back to sitebase
+        self.redirect(media_locale['SITEBASE'])
 
 
 class AuthLoginHandler(BaseHandler):


amgut/handlers/auth_handlers.py
```diff
@@ -42,7 +42,7 @@ def post(self):
             ag_login_id, skid, kitinfo['password'],
             kitinfo['swabs_per_kit'], kitinfo['verification_code'],
             printresults)
-        if success == -1:
+        if success != 1:
```
wasade: inverting the logic here?

amgut/handlers/auth_handlers.py
```diff
@@ -42,7 +42,7 @@ def post(self):
             ag_login_id, skid, kitinfo['password'],
```
wasade: this is tangential to the PR, but, possible to have the password get encrypted if necessary? see here for how to do it:

https://github.com/biocore/american-gut-web/blob/master/scripts/ag#L104

Not _ideal_, but it would work to add the kit then, if `kitinfo['password']` does not identify, issue the method call to migrate the kit password. What do you think?

amgut/handlers/auth_handlers.py
squirrelo: I prefer to have it go to error case in as wide a range of situations as possible. This way if we don't get the exact response we want it sends an error.

wasade: The method can only return `1` and `-1`. Note, the method is a bit [concerning](https://github.com/biocore/american-gut-web/blob/master/scripts/ag#L104), any idea why there is a `commit` in the `except`?

squirrelo: Future-proofing mostly. I agree the method is concerning, but cleanup of the ag_data_access methods should probably happen as another pull request.

squirrelo:
I can put in another pull request to do that shortly.

amgut/handlers/auth_handlers.py
squirrelo: I think that would work nicely, since it catches any issues with passwords not being migrated silently for the user.

wasade: Great, thanks!

wasade:
Awesome!

wasade:
I'm :+1: on this. @adamrp, @josenavas, @beaugunderson or @ElDeveloper, able to review?

beaugunderson:
Logic looks :thumbsup: to me, unsure how to test locally. :)

wasade:
Ya... that's part of the issue. We've got a sporadic 500 error coming that we can't figure out how to recreate locally. It might be part of the redirect, so we just wanted to get that corrected. 

I'm fine with merge unless there are any voices of dissent?

squirrelo:
I figured out what the issue is in #374 so closing this one in favor of that. It's a registration issue.

PR #374, opened 2015-02-24
Add bcrypt check and fix to kit registration

squirrelo:
Does what it says on the tin.

### DIFF
--- a/amgut/handlers/auth_handlers.py
+++ b/amgut/handlers/auth_handlers.py
@@ -49,6 +49,8 @@ class AuthRegisterHandoutHandler(BaseHandler):
             ag_login_id, skid, password, kitinfo['swabs_per_kit'],
             kitinfo['verification_code'], printresults)
         if success == -1:
+            # log them out and send to error page
+            self.clear_cookie("skid")
             self.redirect(media_locale['SITEBASE'] + '/db_error/?err=regkit')
             return
 
@@ -60,6 +62,8 @@ class AuthRegisterHandoutHandler(BaseHandler):
             barcode = row[0]
             success = ag_data.addAGBarcode(ag_kit_id, barcode)
             if success == -1:
+                # log them out and send to error page
+                self.clear_cookie("skid")
                 self.redirect((media_locale['SITEBASE'] +
                                '/db_error/?err=regbarcode'))
                 return
@@ -79,10 +83,10 @@ class AuthRegisterHandoutHandler(BaseHandler):
                        sender=media_locale['HELP_EMAIL'])
         except:
             result = media_locale['EMAIL_ERROR']
-            self.render('no_auth_help_request.html', skid=skid, result=result)
+            self.render('help_request.html', skid=skid, result=result)
             return
 
-        self.redirect(media_locale['SITEBASE'])
+        self.redirect(media_locale['SITEBASE'] + "/authed/portal/")
 
 
 class AuthLoginHandler(BaseHandler):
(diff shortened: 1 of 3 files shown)


wasade:
Looks good. Just to check, did you test locally?

squirrelo:
don't know how honestly.

wasade:
Manually add a kit to the handouts table

squirrelo:
Just did that and it fails. The `ag_update_kit_password` function only updates the ag_kit table, not ag_handout_kits. This seems to be an issue. Not sure how we want to proceed.

amgut/handlers/auth_handlers.py
```diff
@@ -38,6 +39,12 @@ def post(self):
         printresults = ag_data.checkPrintResults(skid)
         if printresults is None:
             printresults = 'n'
+
+        # make sure kit has bcrypt encrypted password, and fix if not
+        if not bcrypt.identify(kitinfo['password']):
```
adamrp: Under what circumstances does this function evaluate to `False`? The docstring on the function (`passlib.hash.bcrypt.identify`) is not helpful, and I couldn't find documentation [on their page](http://pythonhosted.org//passlib/lib/passlib.hash.bcrypt.html?highlight=bcrypt#passlib.hash.bcrypt).

wasade: @beaugunderson, can you comment when you have a moment please?

beaugunderson: `bcrypt.identify` returns False if the argument is not a bcrypt hash; the docs are [here](http://pythonhosted.org//passlib/password_hash_api.html?highlight=identify#passlib.ifc.PasswordHash.identify) (it's a member of all of the different hash modules so they just documented it in one place)

adamrp: Aha! Thanks @beaugunderson

squirrelo:
This now logs the user out if you can't register the kit, which was probably what was causing the error. User was logged in but kit errored on registration so there was a weird logged in but no kit associated limbo state.

One major issue now is that, after bcrypt encrypting the password and registering the kit, I can't login to the kit. It just has me try and register it again.

adamrp:
Do we know why the kit errored on registration? Figuring that out would prevent the liminal state, if I understand correctly.

squirrelo:
No, because that render error was covering up the other error. Hopefully with this fix we can see the original error.

squirrelo:
Ok, found the error and fixed it. Should be good for testing on someone elses computer for verification.

wasade:
Thanks @squirrelo. Sorry to drag this out a bit further, but we think it makes more sense to also force the handout kit passwords to be encrypted. Have a separate PR going in (#379) that takes care of that. This makes the identity check redundant

amgut/templates/no_auth_help_request.html
```diff
@@ -0,0 +1,40 @@
+{% extends no_auth_sitebase.html %}
```
wasade: Can this template be reached? Not seeing it in any render calls

squirrelo: It's the base template for the site when the user is logged out, so nothing will call it directly.

squirrelo:
Might be good to have both, just to catch stragglers and have redundancies built in. Taking the [Chaos Monkey](http://blog.codinghorror.com/working-with-the-chaos-monkey/) approach.

wasade:
Chaos Monkey randomly breaks things, doesn't do excessive checks... not
necessarily opposed to keeping it but it shouldn't be necessary.

PR #375, opened 2015-02-24
Fixes for authentication redirects and Open Humans integration

beaugunderson:
In addition to this code the setting `BASE_URL` also needs to be set to:

```
https://microbio.me/AmericanGut/
```

if it's not already. :)

I've tested this locally by proxying HTTP requests to `http://localhost/AmericanGut/*` to a running instance of `american-gut-web` with its `SITEBASE` set to `/AmericanGut`.

### DIFF
--- a/amgut/handlers/open_humans.py
+++ b/amgut/handlers/open_humans.py
@@ -75,7 +75,7 @@ class OpenHumansLoginHandler(BaseHandler, OpenHumansMixin):
     _API_URL = urljoin(AMGUT_CONFIG.open_humans_base_url, '/api')
 
     _OAUTH_REDIRECT_URL = urljoin(AMGUT_CONFIG.base_url,
-                                  '/authed/connect/open-humans/')
+                                  'authed/connect/open-humans/')
 
     _OAUTH_AUTHORIZE_URL = urljoin(AMGUT_CONFIG.open_humans_base_url,
                                    '/oauth2/authorize/')
--- a/amgut/handlers/base_handlers.py
+++ b/amgut/handlers/base_handlers.py
@@ -2,10 +2,10 @@ import logging
 
 from tornado.web import RequestHandler, StaticFileHandler
 
+from amgut import media_locale, text_locale
 from amgut.connections import ag_data
 from amgut.lib.config_manager import AMGUT_CONFIG
 from amgut.lib.mail import send_email
-from amgut import text_locale
 
 
 class BaseHandler(RequestHandler):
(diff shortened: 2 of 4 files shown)


beaugunderson:
cc @wasade, @madprime

beaugunderson:
Also, rather than fiddling with setting up nginx or something I just wrote a tiny node proxy for the testing; it may come in handy if someone else wants to replicate my test:

```js
'use strict';

var express = require('express');
var httpProxy = require('http-proxy');

var americanGut = httpProxy.createProxyServer();

var app = express();

app.all('/AmericanGut/*', function (req, res) {
  req.url = req.url.replace(/^\/AmericanGut/, '');

  console.log('url', req.url);

  americanGut.web(req, res, {
    target: 'http://localhost:8001'
  });
});

app.listen(8011);
```

amgut/handlers/open_humans.py
```diff
@@ -75,7 +75,7 @@ class OpenHumansLoginHandler(BaseHandler, OpenHumansMixin):
     _API_URL = urljoin(AMGUT_CONFIG.open_humans_base_url, '/api')
 
     _OAUTH_REDIRECT_URL = urljoin(AMGUT_CONFIG.base_url,
-                                  '/authed/connect/open-humans/')
+                                  'authed/connect/open-humans/')
```
wasade: Wont this not work with the overloaded `redirect`?

wasade:
One comment, :+1: otherwise. @adamrp, can you review as well when you have a moment?

amgut/handlers/open_humans.py
beaugunderson: The overloaded `redirect` only works under this condition:

- the URL starts with a `/`
- the URL doesn't start with SITEBASE

In this case, `AMGUT_CONFIG.base_url` should be `https://microbio.me/AmericanGut/`, which doesn't start with a slash, so the overloaded redirect doesn't affect it (or any other absolute URL). :)

beaugunderson: (The reason for that one character change is that later arguments `urljoin` will become absolute to a given domain if they contain a forward slash... meaning that:

`urljoin('http://foo.com/abc/', '/def/')` yields `http://foo.com/def/`. This worked in my testing environment because I had no `SITEBASE` set.)

wasade: ah! nice, thanks

adamrp:
:+1:

PR #378, opened 2015-02-25
Updated shipping address

wasade:
Reflects UCSD. British Gut does not need to be updated. Example of render is below.
screen shot 2015-02-24 at 5 57 33 pm

@elainewolfe, does this look correct? 

@ElDeveloper @adamrp, can you merge please assuming @elainewolfe gives the okay?


elainewolfe:
This looks fine.

PR #390, opened 2015-02-26
DOC: Updating the README

mortonjt:
That was embarrassing - had some git problems that needed to be addressed

So this PR is only added a couple of minor changes
* Updating README
* Adding in dbname parameter. This will also users with a specified username to create test databases through the ag script

### DIFF
--- a/scripts/ag
+++ b/scripts/ag
@@ -277,7 +277,8 @@ def patch():
 def _create_database(force=False):
     # Connect to the postgres server
     try:
-        conn = connect(user=AMGUT_CONFIG.user, password=AMGUT_CONFIG.password,
+        conn = connect(dbname='postgres',
+                       user=AMGUT_CONFIG.user, password=AMGUT_CONFIG.password,
                        host=AMGUT_CONFIG.host, port=AMGUT_CONFIG.port)
     except OperationalError as e:
         raise OperationalError("Cannot connect to the server, error is %s" %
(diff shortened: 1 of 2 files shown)


scripts/ag
```diff
@@ -277,7 +277,8 @@ def patch():
 def _create_database(force=False):
     # Connect to the postgres server
     try:
-        conn = connect(user=AMGUT_CONFIG.user, password=AMGUT_CONFIG.password,
+        conn = connect(dbname='postgres',
```
wasade: shouldn't this be `dbname=AMGUT_CONFIG.database`?

mortonjt: I don't think that works - below is the error message when I make your suggested change

```
psycopg2.OperationalError: Cannot connect to the server, error is FATAL:  database "ag_testing2" does not exist
```

This is what I think is happening

psql client needs to connect to some database before it makes any changes, so it needs to connect to an existing database.  The above error is thrown because the `AMGUT_CONFIG.database` hasn't been created yet, so the psql client cannot connect to it. Since the `postgres` database is created by default, this is a good database to connect to before creating a new database.

wasade: ahhhh yes, the chicken and egg problem. makes sense, thanks!

wasade:
:+1:

wasade:
@adamrp or @squirrelo, can this be merged?

squirrelo:
:+1:

PR #393, opened 2015-03-02
Single transaction register system

squirrelo:
This adds the queue system from qiita to this repository and then uses it to register kits. It also disables the submit button on the register page once it's clicked, so we don't get the multiple registration eror we've been seeing.


amgut/handlers/auth_handlers.py
```diff
@@ -23,6 +23,8 @@ def get(self):
     @authenticated
     def post(self):
         skid = self.current_user
+        # log user out for register process
+        self.clear_cookie("skid")
         tl = text_locale['handlers']
         info = {}
         for info_column in ("email", "participantname", "address", "city",
```
wasade: we should not allow `participantname` to be `None`. This loop is getting default values, but aren't these fields required?

amgut/handlers/auth_handlers.py
```diff
@@ -23,6 +23,8 @@ def get(self):
     @authenticated
     def post(self):
         skid = self.current_user
+        # log user out for register process
+        self.clear_cookie("skid")
```
wasade: Would it be better to just disable the button as this might open the door to a race condition, where the user's kit is being registered, they get frustrated, and log back in themselves, meaning they either get back to the registration window (if they beat the adding the handout kit) or they get logged in successfully. The former would be a bad situation to encounter

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,68 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
                                     refunded, withdrawn])
         self.connection.commit()
 
+    def registerHandoutKit(self, ag_login_id, supplied_kit_id):
+        """
+        Returns
+        -------
+        int
+            1:  success
```
wasade: please just use `True`, `False`. I don't know why 1 and -1 are in use.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,68 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
...
+
+        # get the barcodes and add them to the queue
+        sql = "SELECT barcode FROM ag_handout_kits WHERE kit_id = %s"
+        for row in conn_handler.execute_fetchall(sql, [supplied_kit_id]):
+            barcode = row[0]
+            conn_handler.add_to_queue(
+                queue, "INSERT INTO barcode (barcode) VALUES (%s)", [barcode])
+            conn_handler.add_to_queue(
+                queue, "INSERT INTO project_barcode (project_id, barcode) "
+                "VALUES (1, %s)", [barcode])
```
wasade: Please do not hardcode IDs

The project ID is tied to the handout kit on handout kit creation, so this step is not necessary.

josenavas:
This PR demonstrates how needed is a release in Glowing-Dangerzone... I'll try to have it by the end of the week, so we can update all the projects that depend on it, and we highly reduce code duplication...

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,68 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
...
+            WHERE kit_id = %s""", [supplied_kit_id])
+
+        try:
+            conn_handler.execute_queue(queue)
+        except psycopg2.IntegrityError:
+            # remove kit since registering barcodes failed
+            conn_handler.execute(
+                "DELETE FROM ag_kit WHERE ag_kit_id = %s", [kit_id])
+
+            logging.exception('Error on skid %s:' % ag_login_id)
```
wasade: Possible to add more detail on the error message?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,68 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
...
+
+        # remove from handout_kits table since now registered
+        conn_handler.add_to_queue(
+            queue, """DELETE FROM ag_handout_kits
+            WHERE kit_id = %s""", [supplied_kit_id])
+
+        try:
+            conn_handler.execute_queue(queue)
+        except psycopg2.IntegrityError:
+            # remove kit since registering barcodes failed
```
wasade: if execution fails, then the queue should rollback, in which case we definitely do not want to delete the handout kit

wasade:
Okay, I'll hold off review then

josenavas:
@wasade Do you have a notes or some kind of cookbook-style instructions to do releases (e.g. biom)? I've never done a release like that so having some instructions will help.

amgut/handlers/auth_handlers.py
squirrelo: This is a loop to go through and get the form values from the registration page. We have checks in place on that page to make sure they aren't none.

amgut/handlers/auth_handlers.py
squirrelo: It does disable the button as well. The logout is for in case it errors, so we don't get that weird limbo state of logged in but no registered kits.

amgut/lib/data_access/ag_data_access.py
squirrelo: Copy paste from original code, but can change it.

amgut/lib/data_access/ag_data_access.py
squirrelo: This is a direct copy from the procedure adding the barcodes. Not sure what to do here in this case then.

amgut/lib/data_access/ag_data_access.py
squirrelo: Using logging.exception in an except should also pass the error along with that message before it.

amgut/lib/data_access/ag_data_access.py
squirrelo: This deletes the kit out of the ag_kit table, not the handout kit table. I need to add the kit there first manually so I have the kit ID to attach barcodes to. Since it's not part of the queue, if something fails I need to manually delete it as well.

wasade:
@josenavas, [more or less](https://github.com/biocore/pyqi/blob/master/pyqi/commands/make_release.py)

amgut/lib/data_access/ag_data_access.py
wasade: God damnit, seriously? Can you link directly to that please?

adamrp: Looking at the schema, it looks like kits aren't tied to projects, but rather individual barcodes. So this might be necessary, unless you're saying that all barcodes in a kit are tied to the same project automatically upon the kit's creation (but I'm not seeing where that is done in the code).

amgut/lib/data_access/ag_data_access.py
wasade: oh, I see.

amgut/lib/data_access/ag_data_access.py
wasade: kk

amgut/lib/data_access/ag_data_access.py
wasade: They are, see [here](http://glab.colorado.edu/knight/labadmin/blob/master/scripts/make_kits_and_insert.py#L295)

amgut/lib/data_access/ag_data_access.py
adamrp: :+1: please change, thanks!

squirrelo:
This pull request's SQLConnectionHandler merges functionality needed for both qiita and amgut, so probably best to have this one in there before release.

amgut/lib/data_access/ag_data_access.py
squirrelo: I think the project is for the lab barcode scanner database, with AG being project 1.

josenavas:
@squirrelo which are the specific functionality needed here that is not in glowing dangerzone? Can you add an issue there?

amgut/lib/data_access/ag_data_access.py
wasade: no, it is so we can differentiate what subproject the barcode is associated with. This is very bad.

squirrelo: @wasade also that link you asked for: [ag_insert_barcode.sql](https://github.com/biocore/american-gut-web/blob/master/amgut/lib/data_access/procedures/ag_insert_barcode.sql)

squirrelo: I think this is also pointing to why procedures for simple insert/select is A Bad Idea (tm).

wasade: it would be the same effect if it was inline sql.

squirrelo:
The procedure call function at the end is needed for now.

josenavas:
Oh I see (It's not part of the PR). Can you add an issue to the gd repo? Thanks!

amgut/lib/data_access/sql_connection.py
```diff
@@ -1,69 +1,134 @@
...
+        specified in the qiita configuration, but not to a specific database.
+        If 'admin_with_database', then a connection will be made to the server
+        and database specified in the qiita config.
+    """
+    def __init__(self, con=None, admin='no_admin'):
+        if admin not in ('no_admin', 'admin_with_database',
+                         'admin_without_database'):
+            raise RuntimeError("admin takes only {'no_admin', "
+                               "'admin_with_database', or "
+                               "'admin_without_database'}")
```
adamrp: The postgres section of the config (ag_config.txt.example) should be updated to contain these options (admin_user, admin_password)

amgut/lib/data_access/ag_data_access.py
squirrelo: Yes, but nobody really looks at the procedures files. If it's in the actual code, more people are likely to look at it and catch the problem.

amgut/lib/data_access/ag_data_access.py
squirrelo: Although if (this)[https://github.com/biocore/glowing-dangerzone/issues/2] issue gets fixed I can insert then use the returned ID all in the same queue.

amgut/lib/data_access/sql_connection.py
adamrp: Oh, and config_manager.py

adamrp:
Other than the hardcoded ID for project, I think this looks good.

squirrelo:
What's the status on this?

adamrp:
Need to fix the hardcoded project ID. @josenavas any update on GD?

squirrelo:
Since the project ID is hardcoded anyway, can we fix that in a separate PR? This current functionality will fix a couple of serious bugs.

wasade:
Which ones?

squirrelo:
The major one is that the registration of a kit can screw up and get caught in limbo between logged in but not registered. That was seen at least a few times and is what necessitated this pull request.

josenavas:
Something that we've discussed in a recent meeting (and I think it's included in the document that we wrote) is that we are going to merge those projects back to qiita, as it is really hard to maintain them, given that Qiita and GD are still on high development. If we want to move it back to qiita, will it be a big deal for this repo to depend on the Qiita code? If it is, I can get GD in a reasonable state once we are done with the Qiita release...

wasade:
@josenavas, 6 weeks ago you said GD would be ready by the end of that week. That didn't happen, and now to compound the situation, it's not going to be made available separate of qiita. From the commit history, it doesn't really look like GD is in high development since there hasn't been a change to the codebase in 5 months, and the commit history on qiita's sql connection is pretty light as well.  

From what I can see, there are hardcoded components and database expectations in qiita that make reuse of the connection handler elsewhere risky without an audit and database changes. 

@squirrelo, given the need for transaction blocks, and the relative simplicity of AGweb, do you think we can just juggle isolation level instead of using queues? Agree re: separate pull request at this point.

josenavas:
Sorry @wasade a lot of things have been pushing. I can work on this at the same time that we are working on the Qiita release (i.e. while reviews on my PR are happening). I did not add this to the TODO list before so it slipped through. Sorry about that. I was just bringing that point up to see if we still want to maintain the project by itself or joining into Qiita as it was decided in a call.

wasade:
Okay, thanks @josenavas. 

@squirrelo, can you comment on queues vs. native transaction block support here?

squirrelo:
I think we can use native transaction blocks here, but it will require a bit of changes on the db interface.

wasade:
What about a method with the following signature:

def transaction_block(self, queries, arguments) -> None

queries are a list of str
arguments are a list of tuple

What that method would do is just set the isolation level, iter over args,
rollback if except otherwise proceed forward. We're in a much cleaner
situation here since we don't need returns (afaik) from these which is the
main issue iirc with qiita

squirrelo:
I should be able to code it as a single query to do everything needed to register a kit. It will be much cleaner that way as well. Also no need for queues.

squirrelo:
Rolled back queue addition and now use one giant SQL statement to register kits.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,58 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
...
+                    INSERT INTO barcode (barcode) VALUES (bc);
+                    INSERT INTO project_barcode (project_id, barcode)
+                        VALUES (1, 'bc');
+                    INSERT  INTO ag_kit_barcodes
+                        (ag_kit_id, barcode, sample_barcode_file)
+                        VALUES (k_id, 'bc', 'bc' || '.jpg');
+                END LOOP;
+                DELETE FROM ag_handout_kits WHERE kit_id = '{1}';
+            END $do$;
+            """.format(ag_login_id, supplied_kit_id, printresults)
```
wasade: `supplied_kit_id` isn't scrubbed here, so I believe this is prone to an injection. Possible to pass this in during the execute call and let psycopg2 take care of the scrubbing?

wasade:
Just a question on the query. Will pulldown and test after seeing a response on it

amgut/lib/data_access/ag_data_access.py
squirrelo: Scrubbed, and a bug fixed because of it.

squirrelo:
Tests were passing, but register didn't have a test. Now it does, so yay for that. Should be good to pull down and test now.

wasade:
:+1:

@squirrelo, do you think this will resolve the procedure issue that @adamrp emailed about last night?

@adamrp or @ElDeveloper, possible for one additional review?

squirrelo:
I think so, sonce we no longer use the procedure so we shouldn't have that issue, and it's all one call.

wasade:
That's excellent

amgut/handlers/auth_handlers.py
```diff
@@ -23,10 +23,14 @@ def get(self):
...
-        info = {}
-        for info_column in ("email", "participantname", "address", "city",
-                            "state", "zip", "country"):
+        info = {
+            "email": self.get_argument("email"),
+            "participantname": self.get_argument("participantname")
+        }
+        for info_column in ("address", "city", "state", "zip", "country"):
             # Make sure that all fields were entered
             info[info_column] = self.get_argument(info_column, None)
```
adamrp: The comment here may be out of date. This does not make sure that all fields were entered, just makes sure that the `info` dict has a value (even if it's `None`) for address, city, state, zip, and country.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,6 +386,58 @@ def updateAGBarcode(self, barcode, ag_kit_id, site_sampled,
...
+                    INSERT INTO barcode (barcode) VALUES (bc);
+                    INSERT INTO project_barcode (project_id, barcode)
+                        VALUES (1, bc);
+                    INSERT  INTO ag_kit_barcodes
+                        (ag_kit_id, barcode, sample_barcode_file)
+                        VALUES (k_id, bc, bc || '.jpg');
+                END LOOP;
+                DELETE FROM ag_handout_kits WHERE kit_id = %s;
+            END $do$;
+            """.format(ag_login_id, printresults)
```
adamrp: I like this query :smiley:

adamrp:
:+1:

wasade:
Rock stars.

PR #395, opened 2015-03-03
Need to use None instead of '' for stored proc

adamrp:
This was causing the webserver to crash when fetching elevation data failed.


wasade:
@adamrp, this is kosher, right? Tempted to just merge...

squirrelo:
Looks good to me.

ElDeveloper:
:+1:

PR #402, opened 2015-03-26
Updated mod1main to reflect nov 25 processing

wasade:
bring mod1 up to nov release


adamrp:
I think the content of the document looks fine. One note (which you and I discussed briefly, @wasade); do we think it would be better to include this document as a TIFF or high-resolution PNG? The PDF is  slow to load and scroll, and maybe a different format would behave better?

Surprisingly, printing the current PDF to a file (i.e., "rasterizing" the PDF) grew the file size fro 18 to 27 MB on my laptop, and was not faster to load or scroll.

adamrp:
actulaly TIFF is probably not good, since the file would likely be enormous.

JWDebelius:
We may want to consider an alternative to the 3D emperor plot. @cuttlefishh had some ideas in November.

wasade:
or we just rasterize it better

ElDeveloper:
I suspect the reason why scrolling is slow, is because the included
images are SVGs or other scalable format. Maybe PNGs should be used
instead? At 320 DPI it would be hard to notice the difference and would
likely make the scrolling more responsive.

adamrp:
@ElDeveloper yes, I think that is the issue (and why it loads so slowly, too). PNG is probably a good bet. I'm going to merge this for now, but I created #403 to investigate alternatives.

PR #404, opened 2015-03-31
Add question_shortname question to question table

adamrp:
Needed for metadata pulldown so that we know what to name the column in the
pulldown TSV


amgut/db/patches/0005.sql
```diff
@@ -0,0 +1,13 @@
...
+it, we cannot easily tell what the column should be called in the pulldown TSV.
+
+- Adam Robbins-Pianka
+*/
+
+ALTER TABLE ag.survey_question ADD question_shortname varchar(100) DEFAULT 'TODO' NOT NULL;
+
+COMMENT ON COLUMN ag.survey_question.question_shortname IS 'A QIIME mapping file-compatible header that describes the question and can be used for metadata pulldown';
+
+ALTER TABLE ag.survey_question ADD CONSTRAINT idx_survey_question_1 UNIQUE ( question_shortname ) ;
```
wasade: can't unique if there's the default. resolving

adamrp:
Let's hold off on merging this until we test and make sure we know how to deal with all response types (SINGLE, MULTIPLE, TEXT, STRING).

wasade:
so it doesn't look like its actually covering those? i suspect not based on
our conversation last night...

adamrp:
Yeah, need to look a little closer still, plan to poke at it after lunch.

Adam

adamrp:
right now, we are only pulling down the single-select (i.e., dropdown menu) selections. So let's hold off for now.

wasade:
@adamrp, this is the patch PR that was discussed in the AG meeting, right?

adamrp:
Yep!

wasade:
I'm :+1:, @squirrelo can you look when you get a sec?

PR #409, opened 2015-04-21
BUG:Clear promoted survey ID on participant delete

### LINKED ISSUE #408: Cannot delete promoted surveys
There is a FK on `promoted_survey_ids` table, need to add a `DELETE` on that table [here](https://github.com/biocore/american-gut-web/blob/master/amgut/lib/data_access/ag_data_access.py#L434) for the appropriate survey_id

adamrp:
Fix #408

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -432,6 +432,9 @@ class AGDataAccess(object):
                    "survey_id = %s")
             curr.execute(sql, [survey_id])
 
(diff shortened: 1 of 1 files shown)


squirrelo:
:+1:, but we should probably look at changing this over to a single query, ala [this code](https://github.com/squirrelo/american-gut-web/blob/queued-register-system/amgut/lib/data_access/ag_data_access.py#L406)

adamrp:
Can you create an issue please and label enhancement?

Issue #410, opened 2015-04-22
Lat/long/elevation information needs to be per-survey


adamrp:
Right now, it's recorded in the `ag_login` table, but that is not specific enough. Since we ask for current zip code on the survey, we should record this information in a a new table
```sql
create table geocode_lookup (zip varchar primary key, latitude float, longitude float, elevation float)
```
Suggested course of action is to create a patch that
- [ ] Creates the `geocode_lookup` table described above
- [ ] Updates the table with the existing information from the `ag_login` table
```sql
insert into geocode_lookup select zip, latitude, longitude, elevation from ag_login
```

Then:
- [ ] Modify [add_geocoding_info](https://github.com/biocore/american-gut-web/blob/master/amgut/lib/data_access/ag_data_access.py#L624) function to pull zip codes from `survey_answers_other` and add them the new table rather than the `ag_login` table.
- [ ] Modify the labadmin handler for update geocoding
- [ ] Run the geocoding update function through labadmin

adamrp:
Is this really done? Isn't lat/long/elevation actually currently per login?

squirrelo:
Yes, it's now done by zipcode in survey.

adamrp:
Right! thanks.

PR #411, opened 2015-04-23
Use zipcodes table instead of info in ag_login

adamrp:
This will address parts of #410

### DIFF
--- a/amgut/db/ag.dbs
+++ b/amgut/db/ag.dbs
@@ -285,10 +285,7 @@
 			<column name="state" type="varchar" length="100" jt="12" />
 			<column name="zip" type="varchar" length="10" jt="12" />
 			<column name="country" type="varchar" length="100" jt="12" />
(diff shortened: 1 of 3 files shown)


adamrp:
dont merge yet, still tinkering

adamrp:
okay, ready for final review and merge

wasade:
Is this going to impact inserts into ag_login?

Should zipcode be varchar? US ones can include a dash, are prefixed by
zeros (eg CT) and the field is used for postal codes in other countries
where the code is not assured to be numeric

adamrp:
> Is this going to impact inserts into ag_login?

Good point. Things may have to be modified there...

> Should zipcode be varchar? US ones can include a dash, are prefixed by
zeros (eg CT) and the field is used for postal codes in other countries
where the code is not assured to be numeric

The zipcode column in the zipcodes table is a varchar, or am I mistaken?

wasade:
ah, sorry, looked incorrectly when reading on my phone

PR #412, opened 2015-04-24
Update ag_kit_barcodes table to have survey_ids

adamrp:
ag_kit_barcodes was lacking survey_ids for barcodes associatd with promoted surveys were lacking survey_id


adamrp:
@wasade points out that this does not need to be a patch :laughing: 

well, at least this PR will live here forever to provide this rationale!

PR #415, opened 2015-05-04
Hide "log sample" if no barcodes remaining

squirrelo:
Quick fix to make it less confusing for users. If no barcodes are available for logging, the option is no longer shown.


amgut/templates/sitebase.html
```diff
@@ -151,7 +152,7 @@
 {% end %}
                         </ul>
                     </li>
-{% if kit_verified %}
+{% if avail_barcodes %}
```
wasade: why not `if kit_verified and avail_barcodes`? or is that redundant?

squirrelo: redundant. To have any available barcodes at lest one kit must be verified.

wasade:
:+1:

PR #419, opened 2015-05-11
BUG: date_signed needs to be cast to be serialized

wasade:
@squirrelo @adamrp, possible for quick review please?


adamrp:
:+1: works

PR #421, opened 2015-05-28
Fix multi-email issue

squirrelo:
This makes it so you only get a consent form email on first pass at the survey when creating a new participant. It will no longer try and email you after editing a survey.

### DIFF
--- a/amgut/handlers/human_survey.py
+++ b/amgut/handlers/human_survey.py
@@ -1,15 +1,74 @@
 from json import dumps, loads
 from collections import defaultdict
+import logging
 
 from wtforms import Form
 from tornado.web import authenticated
 from tornado.escape import url_escape
 
-from amgut import media_locale
+from amgut import media_locale, text_locale
 from amgut.connections import ag_data, redis
 from amgut.handlers.base_handlers import BaseHandler
 from amgut.lib.util import store_survey
 from amgut.lib.survey_supp import primary_human_survey
+from amgut.lib.mail import send_email
+
+
+def build_consent_form(consent_info):
+    tl = text_locale['new_participant.html']
+    # email out the consent form
+    if consent_info['age_range'] == '0-6':
+        message = ("%s<p>%s: %s</p><p>%s: %s</p><p>%s: %s</p>"
+                   "<p>%s: %s</p><p>%s: %s</p><p>%s: %s</p>") %\
+            (tl['CONSENT_YOUR_CHILD'],
+             tl['PARTICIPANT_NAME'], consent_info['participant_name'],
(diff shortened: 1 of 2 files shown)


wasade:
:+1:

@adamrp, able for quick review? I discussed with @squirrelo and the logic seems correct (will only email on new survey, at the end of the survey, etc)

amgut/handlers/human_survey.py
```diff
@@ -1,15 +1,74 @@
...
 from amgut.connections import ag_data, redis
 from amgut.handlers.base_handlers import BaseHandler
 from amgut.lib.util import store_survey
 from amgut.lib.survey_supp import primary_human_survey
+from amgut.lib.mail import send_email
+
+
+def build_consent_form(consent_info):
+    tl = text_locale['new_participant.html']
+    # email out the consent form
```
adamrp: Minor, but this comment seems misplaced

amgut/handlers/human_survey.py
```diff
@@ -1,15 +1,74 @@
...
+             tl['DATE_SIGNED'], str(consent_info['date_signed']))
+
+    elif consent_info['age_range'] == '18-plus':
+        message = "%s<p>%s: %s</p><p>%s: %s</p><p>%s: %s</p>" %\
+            (tl['CONSENT_18'],
+             tl['PARTICIPANT_NAME'], consent_info['participant_name'],
+             tl['PARTICIPANT_EMAIL'], consent_info['participant_email'],
+             tl['DATE_SIGNED'], str(consent_info['date_signed']))
+
+    else:
```
adamrp: Currently, users can only add one survey per consent so this shouldn't happen, right? Of course, if/when multiple surveys are allowed per consent, this *would* actually happen if someone added a new survey to an existing "old" consent.

adamrp:
Slightly misleading code comment should be (re)moved. Other than that, this looks good, as long as we are not trying to set the code up to accept multiple surveys per consent.

wasade:
that's a bit of a larger issue though anyway, so I think getting this in
now, and tackling multiple surveys per consent latter makes sense?

adamrp:
yep, agreed! Just mentioning it here so that we don't completely forget about it, since it's been on our radar for a while. I'll merge when the comment on line 19 is removed.

squirrelo:
Fixed the comment.

adamrp:
Thanks!

amgut/handlers/human_survey.py
squirrelo: Yeah, I have it there as a catch-all to figure out if things are going wrong, like in the case of catching the multi-email issue.

PR #423, opened 2015-06-15
Check barcode access by login, not kit

squirrelo:
Changes SQL in a small way to check by ag_login_id, not supplied_kit_id


wasade:
none of the calls to `check_access` need to be modified?

squirrelo:
not that I know of. All this does is convert the already passed SKID to the login ID and uses that instead.

adamrp:
This makes sense to me, but I haven't tested it.

PR #425, opened 2015-06-29
Fix registration again

squirrelo:
This does some major changes:
* Completely decoupled login and registration. Now, if a user tries to log in with a non-registered kit, they will get a message informing them to register.
* Added `Register Kit` link to no_auth_sitebase.html
* Registration now works in non-authenticated state, then logs the user in once registration has successfully happened.
* Added all appropriate text to locale dicts.

### DIFF
--- a/amgut/handlers/auth_handlers.py
+++ b/amgut/handlers/auth_handlers.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 from tornado.web import authenticated, HTTPError
-from tornado.escape import json_encode
+from tornado.escape import json_encode, url_escape
 import logging
 
 from amgut.connections import ag_data
@@ -34,7 +34,7 @@ class AuthRegisterHandoutHandler(AuthBasehandler):
         if not is_handout:
             tl = text_locale['handlers']
             self.redirect(media_locale['SITEBASE'] +
-                          "/?loginerror=" + tl['INVALID_KITID'])
+                          "/?loginerror=" + url_escape(tl['INVALID_KITID']))
             return
 
         # Register handout
--- a/amgut/handlers/auth_handlers.py
+++ b/amgut/handlers/auth_handlers.py
@@ -22,9 +22,8 @@ class AuthBasehandler(BaseHandler):
 class AuthRegisterHandoutHandler(AuthBasehandler):
     """User Creation"""
     def get(self):
-        skid = self.get_argument("skid").strip()
         latlong_db = ag_data.getMapMarkers()
-        self.render("register_user.html", skid=skid,
+        self.render("register_user.html",
                     latlongs_db=latlong_db, loginerror='')
 
     def post(self):
@@ -36,7 +35,6 @@ class AuthRegisterHandoutHandler(AuthBasehandler):
             tl = text_locale['handlers']
             self.redirect(media_locale['SITEBASE'] +
                           "/?loginerror=" + tl['INVALID_KITID'])
-            print ">>>REDIRECT", media_locale['SITEBASE'] + "/?loginerror=" + tl['INVALID_KITID']
             return
 
         # Register handout
@@ -94,8 +92,8 @@ class AuthLoginHandler(AuthBasehandler):
         is_handout = ag_data.handoutCheck(skid, password)
         if is_handout:
             # have them register themselves
-            self.redirect(media_locale['SITEBASE'] + '/auth/register/?skid=%s'
-                          % skid)
+            self.redirect(media_locale['SITEBASE'] + '/?loginerror=' +
+                          tl['REGISTER_KIT'])
             return
 
         login = ag_data.authenticateWebAppUser(skid, password)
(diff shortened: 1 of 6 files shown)


amgut/templates/register_user.html
```diff
@@ -24,8 +24,12 @@
         address: "required",
         city: "required",
         state: "required",
-        zip: "required",
-        country: "required"
+        zip: {"required": true,
+              "maxlength": 10
```
wasade: ?

squirrelo: Apparently the max length allowed in the DB is 10 for zipcode.

wasade:
I didn't get any feedback about an incorrect password when registering. Is that possible to add please? 

Also, registration takes a _long_ time and eats a huge amount of CPU. Investigating

squirrelo:
Per our hangouts conversation, both fixed.

amgut/handlers/auth_handlers.py
```diff
@@ -9,22 +9,35 @@
...
-        # log user out for register process
-        self.clear_cookie("skid")
+        # Check handout
+        skid = self.get_argument("kit_id").strip()
+        password = self.get_argument("password")
+        is_handout = ag_data.handoutCheck(skid, password)
+        if not is_handout:
+            tl = text_locale['handlers']
+            self.redirect(media_locale['SITEBASE'] +
+                          "/?loginerror=" + tl['INVALID_KITID'])
```
adamrp: should this text be URL escaped?

amgut/handlers/auth_handlers.py
```diff
@@ -44,8 +57,8 @@ def post(self):
             self.redirect(media_locale['SITEBASE'] + '/db_error/?err=regkit')
```
adamrp: wasn't part of this PR, but this error should probably be put in the text locale, I think

amgut/handlers/auth_handlers.py
```diff
@@ -66,12 +79,23 @@ def post(self):
             logging.exception('Error on skid %s:' % skid)
 
 
-class AuthLoginHandler(BaseHandler):
+class AuthLoginHandler(AuthBasehandler):
     """user login, no page necessary"""
+    def get(self, *args, **kwargs):
```
adamrp: why are `args` and `kwargs` needed here?

adamrp: Oh, maybe they will be automatically passed to the `redirect`?

amgut/lib/locale_data/american_gut.py
```diff
@@ -598,7 +602,7 @@
     'ENTER_STATE': 'Please enter your state',
     'ENTER_ZIP': 'Please enter your zip',
     'ENTER_COUNTRY': 'Please enter your country',
-    'REQUIRED_ZIP': 'Your zip must consist of at least 5 characters',
+    'REQUIRED_ZIP': 'Your zip must be 10 or fewer characters',
```
adamrp: Where is this requirement implemented? It does not appear to be enforced in the database schema definition anywhere. The only limitation I can find is on the `ag_map_markers` table where it is limited to 20.

amgut/templates/register_user.html
adamrp: really? I could not find this (looking at the HTML documentation of the schema; I don't have access to the live system at this time)

adamrp:
A few minor questions, but code looks good overall. One general comment, though, is that it seems weird to me to maintain two different localizations (AG and BG) unless we are *actually* going to ask someone who is more familiar with British dialect to review the additions/"translations" -- otherwise, we are just duplicating verbatim the American English text into the British English locale for effectively no reason (and possibly incorrectly)

amgut/handlers/auth_handlers.py
wasade: They get passed in by tornado if they're in the URL, so you get a `TypeError` due to an incorrect number of args

amgut/lib/locale_data/american_gut.py
wasade: @squirrelo indicated it was currently the width of the db field. This is fine AFAIK for US zip codes

amgut/templates/register_user.html
squirrelo: live system yelled at me for 10, so I assume the schema documentation is wrong then.

amgut/handlers/auth_handlers.py
squirrelo: probably. Fixed.

amgut/handlers/auth_handlers.py
squirrelo: the regkit value is actually a code for what error to show on the page, so it is text-localed.

squirrelo:
Yeah, I agree it's silly in this specific case, but for any actually different language (probably should have spanish right now) it makes sense.

squirrelo:
Ready for final review/merge once duplicate barcodes taken care of

adamrp:
@squirrelo did you take care of the duplicate barcodes? if so, I think this can be merged.

squirrelo:
I can make that a patch?

squirrelo:
Cleaned

PR #426, opened 2015-06-29
Add source info to logging

squirrelo:
This adds a "[NAME] - Human Source" and "[NAME] - Animal Source" to the logging to stop people being confused and not realizing their name corresponds to a human sample.


wasade:
Is this a present issue?

squirrelo:
Yes, was discussing and it came up as something users seem to have trouble with. Instead of looking for their name, they are looking for the specific words `Human Source` and can't quite put together that if it has their name, it is a human source.

PR #428, opened 2015-07-06
Fix handouts table - NEEDS REVIEW

squirrelo:
This breaks up the handout kits into two tables: one for handout kit info and one for barcodes attached to the handout. This fixes the issue we've seen previously, where each row in the `ag_handout_kits` table for a given kit is exactly the same except for the barcode. this breakout means that each kit now has a single entry in that table, and the deletion of a kit from that table cascades down and deletes the barcodes from the second table. This is a much cleaner representation, as it allows us to properly represent kits with multiple barcodes.

### DIFF
--- a/amgut/db/patches/0009.sql
+++ b/amgut/db/patches/0009.sql
@@ -2,11 +2,11 @@
 ALTER TABLE ag.ag_handout_kits ADD created_on timestamp DEFAULT current_timestamp NOT NULL;
 
 -- Create new ag_handout_barcodes table
-CREATE TABLE ag.ag_handout_barcodes ( 
+CREATE TABLE ag.ag_handout_barcodes (
 	kit_id               varchar  NOT NULL,
-	barcode              varchar(9)  NOT NULL UNIQUE,
+	barcode              varchar(9)  NOT NULL,
 	sample_barcode_file  varchar(13),
-	CONSTRAINT idx_ag_handout_barcodes PRIMARY KEY ( kit_id, barcode )
+	CONSTRAINT idx_ag_handout_barcodes PRIMARY KEY ( barcode )
  );
 
 CREATE INDEX idx_ag_handout_barcodes_0 ON ag.ag_handout_barcodes ( barcode );
@@ -19,7 +19,7 @@ ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes FOREIGN
 INSERT INTO ag.ag_handout_barcodes (kit_id, barcode, sample_barcode_file)
 SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
 
--- Delete duplicate handout rows and unneded columns from handouts table
+-- Delete duplicate handout rows and unneeded columns from handouts table
 DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
 
 ALTER TABLE ag.ag_handout_kits DROP COLUMN barcode, DROP COLUMN sample_barcode_file;
--- a/amgut/db/patches/0009.sql
+++ b/amgut/db/patches/0009.sql
@@ -4,7 +4,7 @@ ALTER TABLE ag.ag_handout_kits ADD created_on timestamp DEFAULT current_timestam
 -- Create new ag_handout_barcodes table
 CREATE TABLE ag.ag_handout_barcodes ( 
 	kit_id               varchar  NOT NULL,
-	barcode              varchar(9)  NOT NULL,
+	barcode              varchar(9)  NOT NULL UNIQUE,
 	sample_barcode_file  varchar(13),
 	CONSTRAINT idx_ag_handout_barcodes PRIMARY KEY ( kit_id, barcode )
  );
--- a/amgut/db/patches/0009.sql
+++ b/amgut/db/patches/0009.sql
@@ -9,8 +9,6 @@ CREATE TABLE ag.ag_handout_barcodes (
 	CONSTRAINT idx_ag_handout_barcodes PRIMARY KEY ( barcode )
  );
 
-CREATE INDEX idx_ag_handout_barcodes_0 ON ag.ag_handout_barcodes ( barcode );
-
 CREATE INDEX idx_ag_handout_barcodes_1 ON ag.ag_handout_barcodes ( kit_id );
 
 ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
@@ -20,7 +18,7 @@ INSERT INTO ag.ag_handout_barcodes (kit_id, barcode, sample_barcode_file)
 SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
 
 -- Delete duplicate handout rows and unneeded columns from handouts table
-DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
+DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits GROUP BY kit_id);
 
 ALTER TABLE ag.ag_handout_kits DROP COLUMN barcode, DROP COLUMN sample_barcode_file;
 
--- a/amgut/db/patches/0009.sql
+++ b/amgut/db/patches/0009.sql
@@ -25,5 +25,5 @@ DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag
 ALTER TABLE ag.ag_handout_kits DROP COLUMN barcode, DROP COLUMN sample_barcode_file;
 
 -- Add unique constraint to the kit_id column and the foreign key for new table
-ALTER TABLE ag.ag_handout_kits ADD UNIQUE (kit_id);
+ALTER TABLE ag.ag_handout_kits ADD PRIMARY KEY (kit_id);
 ALTER TABLE ag.handout_barcode ADD CONSTRAINT fk_handout_barcode_0 FOREIGN KEY ( kit_id ) REFERENCES ag.ag_handout_kits( kit_id ) ON DELETE CASCADE ON UPDATE CASCADE;
--- a/amgut/db/patches/0009.sql
+++ b/amgut/db/patches/0009.sql
@@ -1,0 +1,29 @@
+-- Add created_on timestamp for handout kits
+ALTER TABLE ag.ag_handout_kits ADD created_on timestamp DEFAULT current_timestamp NOT NULL;
+
+-- Create new handout_barcode table
+CREATE TABLE ag.handout_barcode ( 
+	kit_id               varchar  NOT NULL,
+	barcode              varchar(9)  NOT NULL,
+	sample_barcode_file  varchar(13),
+	CONSTRAINT idx_handout_barcode PRIMARY KEY ( kit_id, barcode )
+ );
+
+CREATE INDEX idx_handout_barcode_0 ON ag.handout_barcode ( barcode );
+
+CREATE INDEX idx_handout_barcode_1 ON ag.handout_barcode ( kit_id );
+
+ALTER TABLE ag.handout_barcode ADD CONSTRAINT fk_handout_barcode FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
+
+-- Copy needed data to new table
+INSERT INTO ag.handout_barcode (kit_id, barcode, sample_barcode_file)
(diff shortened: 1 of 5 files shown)


squirrelo:
This is ready for review

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE INDEX idx_handout_barcode_1 ON ag.handout_barcode ( kit_id );
+
+ALTER TABLE ag.handout_barcode ADD CONSTRAINT fk_handout_barcode FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
+
+-- Copy needed data to new table
+INSERT INTO ag.handout_barcode (kit_id, barcode, sample_barcode_file)
+SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
+
+-- Delete duplicate handout rows and unneded columns from handouts table
+DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
```
wasade: it looks like the barcode isn't necessary then, right? why not just `alter table ag.ag_handout_kits drop column barcode`?

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+
+ALTER TABLE ag.ag_handout_kits DROP COLUMN barcode, DROP COLUMN sample_barcode_file;
```
wasade: heh...

squirrelo: :bangbang:

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- Add unique constraint to the kit_id column and the foreign key for new table
+ALTER TABLE ag.ag_handout_kits ADD PRIMARY KEY (kit_id);
+ALTER TABLE ag.handout_barcode ADD CONSTRAINT fk_handout_barcode_0 FOREIGN KEY ( kit_id ) REFERENCES ag.ag_handout_kits( kit_id ) ON DELETE CASCADE ON UPDATE CASCADE;
```
wasade: Why the cascades?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -1222,7 +1228,7 @@ def get_barcode_results(self, supplied_kit_id):
         return [dict(zip(col_names, row)) for row in results]
 
     def get_barcodes_from_handout_kit(self, supplied_kit_id):
-        sql = "select barcode from ag_handout_kits where kit_id = %s"
+        sql = "select barcode from handout_barcode where kit_id = %s"
```
wasade: no schema?

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
+-- Add created_on timestamp for handout kits
+ALTER TABLE ag.ag_handout_kits ADD created_on timestamp DEFAULT current_timestamp NOT NULL;
+
+-- Create new handout_barcode table
+CREATE TABLE ag.handout_barcode (
```
wasade: while the naming structure is redundant and its use of plurals is not great, but it is better to be consistent -- could this be changed to `ag_handout_barcodes`?

wasade:
Just a few comments, but appears to be good otherwise

amgut/db/patches/0009.sql
squirrelo: allows you to delete the handout kit from the one table and that cascades the deletion of the barcodes associated with it.

amgut/lib/data_access/ag_data_access.py
squirrelo: do we need it? we have ag added to the search_path.

amgut/db/patches/0009.sql
squirrelo: changed, although yeah the naming structure is bad and needs fixing if possible.

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
+-- Add created_on timestamp for handout kits
+ALTER TABLE ag.ag_handout_kits ADD created_on timestamp DEFAULT current_timestamp NOT NULL;
+
+-- Create new ag_handout_barcodes table
+CREATE TABLE ag.ag_handout_barcodes ( 
+	kit_id               varchar  NOT NULL,
+	barcode              varchar(9)  NOT NULL UNIQUE,
+	sample_barcode_file  varchar(13),
+	CONSTRAINT idx_ag_handout_barcodes PRIMARY KEY ( kit_id, barcode )
```
adamrp: You could make `barcode` the primary key (principle of minimality)

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE INDEX idx_ag_handout_barcodes_1 ON ag.ag_handout_barcodes ( kit_id );
+
+ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
+
+-- Copy needed data to new table
+INSERT INTO ag.ag_handout_barcodes (kit_id, barcode, sample_barcode_file)
+SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
+
+-- Delete duplicate handout rows and unneded columns from handouts table
+DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
```
adamrp: If I'm reading this correctly, it will delete all but one row from `ag.ag_handout_kits`. Is that really the intent?

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE INDEX idx_ag_handout_barcodes_1 ON ag.ag_handout_barcodes ( kit_id );
+
+ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
+
+-- Copy needed data to new table
+INSERT INTO ag.ag_handout_barcodes (kit_id, barcode, sample_barcode_file)
+SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
+
+-- Delete duplicate handout rows and unneded columns from handouts table
```
adamrp: `unneded` -> `unneeded`

amgut/db/populate_test.sql
```diff
@@ -77,10 +77,10 @@ insert into project_barcode (project_id, barcode) values (1, '000010860'), (1, '
...
         ('ded369ec-7d0f-4d8c-e040-8a80115d6742','d8592c74-7da2-2135-e040-8a80115d6401','000006616',19,'000006616.jpg','','Nasal mucus','08/12/2014',
             'MAKE SURE THIS IS BEING PASSED','12:05 PM','','','','','','','','','','');
-insert into ag_handout_kits (kit_id, password, barcode, verification_code, swabs_per_kit, print_results)
-    values('test_ha', '$2a$12$GfYI6sxlVGTamDJ3FEQzMuQ46eSV6YJ.DxUxiZvKJaAn1NJtTDtFS', '000000002', '5678', '3', 'n'),
-          ('test_ha', '$2a$12$GfYI6sxlVGTamDJ3FEQzMuQ46eSV6YJ.DxUxiZvKJaAn1NJtTDtFS', '000000003', '5678', '3', 'n'),
-          ('test_ha', '$2a$12$GfYI6sxlVGTamDJ3FEQzMuQ46eSV6YJ.DxUxiZvKJaAn1NJtTDtFS', '000000004', '5678', '3', 'n');
+insert into ag_handout_kits (kit_id, password, verification_code, swabs_per_kit, print_results)
+    values('test_ha', '$2a$12$GfYI6sxlVGTamDJ3FEQzMuQ46eSV6YJ.DxUxiZvKJaAn1NJtTDtFS', '5678', '3', 'n');
+insert into ag_handout_barcodes (kit_id, barcode)
+    values ('test_ha', '000000002'),('test_ha', '000000003'),('test_ha', '000000004');
```
adamrp: These changes really don't break any tests? That's surprising, and might indicate that our code test coverage should be improved.

For example, the `distinct` can be removed from `amgut.lib.data_access.ag_data_access.handoutCheck`. I suspect there are other changes that should be made, too, but will leave it to @squirrelo to investigate further. I gave it a quick look, and it appears that most of the functions that operate on the `ag_handout_kits` are untested, which makes me concerned that unforeseen errors will arise in the future.

adamrp: Somehow I missed the changes to ag_data_access.py on my first pass here -- I see that you made several changes there. However it would still be good to have tests.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -944,7 +950,7 @@ def addParticipantException(self, ag_login_id, participant_name):
 
     def handoutCheck(self, username, password):
         cursor = self.get_cursor()
-        cursor.execute("""SELECT distinct(password)
+        cursor.execute("""SELECT DISTINCT (password)
```
adamrp: I *think* the new ag_handout_kits table will have only one row per kit ID, which means only one password, which means that the `distinct` here should be unnecessary. If all of my assumptions are correct and things break when the `distinct` keyword is removed, it's probably a cause for concern.

adamrp:
Done with first round of review, had some comments/questions

amgut/db/patches/0009.sql
squirrelo: Yes. There is one row per barcode currently, with all other information duplicated if there are multiple barcodes per handout kit, hence the function of this pull request. This allows us to have a single entry per handout kit.

amgut/lib/data_access/ag_data_access.py
squirrelo: Yeah, that was a thing from the old SQL that wasn't removed. Good catch.

amgut/db/patches/0009.sql
squirrelo: True. Will do.

amgut/db/populate_test.sql
squirrelo: There are tests, but they haven't been updated for other things and don't pass on master either. We should probably set up travis and coveralls on this repo.

squirrelo:
ready for another review

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE INDEX idx_ag_handout_barcodes_1 ON ag.ag_handout_barcodes ( kit_id );
+
+ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes FOREIGN KEY ( barcode ) REFERENCES ag.barcode( barcode );
+
+-- Copy needed data to new table
+INSERT INTO ag.ag_handout_barcodes (kit_id, barcode, sample_barcode_file)
+SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
+
+-- Delete duplicate handout rows and unneeded columns from handouts table
+DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
```
adamrp: I don't think this statement does what you want. It will literally leave exactly one row in the table *total* (i.e., the row corresponding to the highest barcode in the system) whereas what I think you want is one row per handout kit. To do the latter, your subquery needs a groupby on kit_id

amgut/db/populate_test.sql
adamrp: This should be a top priority, IMHO, and I created #434

adamrp:
Still not convinced about the `DELETE` from ag_handout_kits (did you test this?), but the rest looks good now. I Created an issue (#434) for Travis.

amgut/db/patches/0009.sql
squirrelo: Oh wow, that is special. Thought I had the groupby in there :disappointed:

wasade: Josh, can you please verify this with a local copy of the production data?

squirrelo:
delete now fixed and tested with multiple kits.

adamrp:
Okay, thanks! :+1:

adamrp:
@squirrelo I think @wasade wants confirmation that the patch looks good on a local copy of what's currently live. After you do that, we can merge.

squirrelo:
Using the backup I made before cleaning the handouts table:
ag_test=# select count(distinct kit_id) from ag.ag_handout_kits; 
 count 
-------
  3612
(1 row)

ag_test=# select count(kit_id) from ag.ag_handout_kits;      
 count 
-------
 11218
(1 row)

`APPLY PATCH`

ag_test=# select count(kit_id) from ag.ag_handout_kits;
 count 
-------
  3612
(1 row)

Verified working.

squirrelo:
Also 

ag_test=# select count(*) from ag.ag_handout_barcodes;
 count 
-------
 11218
(1 row)

PR #429, opened 2015-07-20
Third party survey tables

squirrelo:
~~Depends on #428~~

This adds two tables to the database, one for storing the external surveys available, and one to store the answers to said surveys in JSON format, on a per survey basis.

### DIFF
--- a/amgut/db/patches/0010.sql
+++ b/amgut/db/patches/0010.sql
@@ -1,16 +1,17 @@
-CREATE TABLE ag.external_survey ( 
+CREATE TABLE ag.external_survey_sources (
 	external_survey_id   bigserial  NOT NULL,
 	external_survey      varchar  NOT NULL,
 	external_survey_description varchar  ,
+	external_survey_url varchar  NOT NULL,
 	CONSTRAINT pk_external_surveys PRIMARY KEY ( external_survey_id )
  );
 
-CREATE TABLE ag.external_survey_answers ( 
+CREATE TABLE ag.external_survey_answers (
 	survey_id            varchar  NOT NULL,
 	external_survey_id   bigint  NOT NULL,
 	pulldown_date        date  NOT NULL,
 	answers              json  NOT NULL,
-	CONSTRAINT pk_third_party_surveys PRIMARY KEY ( survey_id )
+	CONSTRAINT pk_third_party_surveys PRIMARY KEY ( survey_id, external_survey_id, pulldown_date )
  );
 
 CREATE INDEX idx_external_survey ON ag.external_survey_answers ( external_survey_id );
@@ -19,4 +20,4 @@ COMMENT ON TABLE ag.external_survey_answers IS 'This stores answers to third par
 
 ALTER TABLE ag.external_survey_answers ADD CONSTRAINT fk_third_party_surveys FOREIGN KEY ( survey_id ) REFERENCES ag.ag_login_surveys( survey_id );
 
-ALTER TABLE ag.external_survey_answers ADD CONSTRAINT fk_external_survey FOREIGN KEY ( external_survey_id ) REFERENCES ag.external_survey( external_survey_id );
+ALTER TABLE ag.external_survey_answers ADD CONSTRAINT fk_external_survey FOREIGN KEY ( external_survey_id ) REFERENCES ag.external_survey_sources( external_survey_id );
--- a/amgut/db/patches/0010.sql
+++ b/amgut/db/patches/0010.sql
@@ -1,0 +1,22 @@
+CREATE TABLE ag.external_survey ( 
+	external_survey_id   bigserial  NOT NULL,
+	external_survey      varchar  NOT NULL,
+	external_survey_description varchar  ,
+	CONSTRAINT pk_external_surveys PRIMARY KEY ( external_survey_id )
+ );
+
+CREATE TABLE ag.external_survey_answers ( 
+	survey_id            varchar  NOT NULL,
+	external_survey_id   bigint  NOT NULL,
+	pulldown_date        date  NOT NULL,
+	answers              json  NOT NULL,
+	CONSTRAINT pk_third_party_surveys PRIMARY KEY ( survey_id )
+ );
+
+CREATE INDEX idx_external_survey ON ag.external_survey_answers ( external_survey_id );
+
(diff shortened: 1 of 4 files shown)


amgut/db/patches/0010.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE ag.external_survey (
```
wasade: Could this be changed to `external_survey_type`?

wasade:
One minor comment, +1 otherwise except for pending comments on #428

amgut/db/patches/0010.sql
squirrelo: I'm not sure type is a good descriptor here, since they are full surveys in and of themselves, with names, not types.

amgut/db/patches/0010.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE ag.external_survey ( 
+	external_survey_id   bigserial  NOT NULL,
+	external_survey      varchar  NOT NULL,
+	external_survey_description varchar  ,
```
adamrp: Might be good to add a timestamp field to this table, since the table is storing the contents of a particular survey that was retrieved at some particular point in time.

adamrp: I think I misunderstood the purpose of this table. If this is just storing the name of the survey and giving it a handle in our database, then I think @wasade is correct and it should be renamed to `external_survey_types` (or something like `external_survey_sources` if you don't like `types`)

amgut/db/patches/0010.sql
```diff
@@ -0,0 +1,22 @@
...
+	CONSTRAINT pk_external_surveys PRIMARY KEY ( external_survey_id )
+ );
+
+CREATE TABLE ag.external_survey_answers ( 
+	survey_id            varchar  NOT NULL,
+	external_survey_id   bigint  NOT NULL,
+	pulldown_date        date  NOT NULL,
+	answers              json  NOT NULL,
+	CONSTRAINT pk_third_party_surveys PRIMARY KEY ( survey_id )
```
adamrp: What if there are multiple external surveys associated with a single american gut survey? I think this primary key needs to be expanded to include the `external_survey_id`.

Additionally, if the plan is to store multiple "versions" of a person's answers to an external survey (retrieved on different dates), then the `pulldown_date` should also be included in the primary key, as well. If that is *not* the plan, I'm confused why this table cannot be combined with the `external_survey` table.

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,29 @@
...
+SELECT kit_id, barcode, sample_barcode_file FROM ag.ag_handout_kits;
+
+-- Delete duplicate handout rows and unneded columns from handouts table
+DELETE FROM ag.ag_handout_kits WHERE barcode NOT IN (SELECT max(barcode) FROM ag.ag_handout_kits);
+
+ALTER TABLE ag.ag_handout_kits DROP COLUMN barcode, DROP COLUMN sample_barcode_file;
+
+-- Add unique constraint to the kit_id column and the foreign key for new table
+ALTER TABLE ag.ag_handout_kits ADD PRIMARY KEY (kit_id);
+ALTER TABLE ag.ag_handout_barcodes ADD CONSTRAINT fk_ag_handout_barcodes_0 FOREIGN KEY ( kit_id ) REFERENCES ag.ag_handout_kits( kit_id ) ON DELETE CASCADE ON UPDATE CASCADE;
```
adamrp: I think this file is unchanged from what I already reviewed in #428. Let me know if I'm mistaken and should review this here, too.

amgut/db/populate_test.sql
```diff
@@ -90,3 +90,7 @@ insert into survey_answers (survey_id, survey_question_id, response)
     values (19, 1, 'Unspecified');
 insert into survey_answers (survey_id, survey_question_id, response)
     values (19, 2, 'Yes');
+insert into external_survey (external_survey, external_survey_description)
+    values ('vioscreen', 'dietary survey');
+insert into external_survey_answers (survey_id, external_survey_id, pulldown_date, answers)
+    values (19, 1, '2015-07-20', '{"chicken":"lots","grains":"some","greens":5}');
```
adamrp: I think this file is unchanged from what I already reviewed in #428. Let me know if I'm mistaken and should review this here, too.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -1319,9 +1325,12 @@ def get_barcode_info_by_kit_id(self, ag_kit_id):
...
-                 or barcode like %s"""
+        sql = """SELECT kit_id, password, barcode, verification_code
+                 FROM ag_handout_kits
+                 JOIN (SELECT kit_id, barcode, sample_barcode_file
+                    FROM ag.ag_handout_barcodes
+                    GROUP BY kit_id, barcode) AS hb USING (kit_id)
+                 WHERE kit_id LIKE %s or barcode LIKE %s"""
         cursor = self.get_cursor()
         liketerm = '%%' + term + '%%'
         cursor.execute(sql, [liketerm, liketerm])
```
adamrp: I think this file is unchanged from what I already reviewed in #428. Let me know if I'm mistaken and should review this here, too.

amgut/db/patches/0010.sql
squirrelo: sources is probably better, since it would also be good to store the URL here as well.

amgut/db/patches/0010.sql
squirrelo: The first combined key idea is probably best. I didn't combine the two tables because it makes it harder to add/remove/query for what external surveys are available to ag users. I originally had it as an enum for external_survey_id to control what surveys there were, but it got cumbersome for adding and querying survey availability.

squirrelo:
ready for another review.

adamrp:
Looks good now, thanks! :+1: after #428 is in.

squirrelo:
Merged in #428 so ready for final review.

wasade:
This creates the structures to store the third party data, but shouldn't this be paired with a mechanism to populate those structures?

squirrelo:
It will be part of labmin, but no current pull request yet.

wasade:
okay, thanks!

PR #431, opened 2015-07-23
Blank participant fix

squirrelo:
Somehow 5 people made it through the survey giving a blank name and email. Not sure how, but htis catches it and has them consent with a proper name and email.

### DIFF
--- a/amgut/lib/locale_data/british_gut.py
+++ b/amgut/lib/locale_data/british_gut.py
@@ -89,6 +89,7 @@ media_locale = {
 _HANDLERS = {
     'PARTICIPANT_EXISTS': 'Participant %s already exists!',
     'SUCCESSFULLY_ADDED': "Successfully added %s!",
+    'MISSING_NAME_EMAIL': 'Missing participant name or email. Please retry, adding all required information.',
     'AUTH_REGISTER_SUBJECT': "%(project_shorthand)s Verification Code" % {'project_shorthand': AMGUT_CONFIG.project_shorthand},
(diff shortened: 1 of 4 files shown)


wasade:
Please update the js validation to ensure the name contains non-whitespace characters as ' ' will pass the validation

squirrelo:
Done, ready for review

amgut/lib/locale_data/british_gut.py
```diff
@@ -89,6 +89,7 @@
 _HANDLERS = {
     'PARTICIPANT_EXISTS': 'Participant %s already exists!',
```
adamrp: Not part of this PR, but I was just noticing that the american and british versions of `_HANDLERS` do not contain exactly the same keys. In particular, the `JUVENILE_CONSENT_EXPECTED` key is missing from this file (not sure about any others). I *would* say that this should lead to bugs when the text locale is assumed to contain exactly the same set of keys, but as far as I can tell the key is never accessed. After verifying that the `JUVENILE_CONSENT_EXPECTED` key is *actually* never accessed, it should probably be removed from the american version. Might also want to verify that, otherwise, the two dicts contain exactly the same keys.

amgut/templates/new_participant.html
```diff
@@ -10,7 +87,7 @@
     }
 
     function relax_parents(box) {
-        var sec_parent = document.getElementById(box.form.id + "_parent_2_name");
+        var sec_parent = $("#" + box.form.id + "_parent_2_name");
```
adamrp: Just curious about the purpose of this change. Also, I like the name of this function :smile:

adamrp:
One nonblocking comment, and one comment about the integrity of the text locales that does not impact this PR. :+1:

amgut/templates/new_participant.html
squirrelo: uses jquery instead of standard javascript. Was done to fix an issue with safari, but was not actually fixed with this, so can revert if wanted. Works either way though.

amgut/lib/locale_data/british_gut.py
squirrelo: Yup, it's a holdover from the old consent forms setup. Removed.

PR #432, opened 2015-07-23
Pet survey

squirrelo:
Pet survey is back, using the new survey style.


rob-knight:
Thanks!

On Jul 23, 2015, at 3:45 PM, Joshua Shorenstein <notifications@github.com>
wrote:

Pet survey is back, using the new survey style.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/biocore/american-gut-web/pull/432
Commit Summary

   - first pass at databse entries for pet survey
   - proper survey_id
   - proper question type
   - move code for general survey use
   - use WTForms for animal survey
   - add unspecified as default
   - simplify form call
   - add survey validation
   - update getAnimalParticipants for new survey setup
   - fix erronius question type
   - store pet survey in DB
   - show humans by survey taken
   - log multiple pets living with participant pet

File Changes

   - *A* amgut/db/patches/0009.sql
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-0>
   (229)
   - *M* amgut/handlers/animal_survey.py
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-1>
   (115)
   - *M* amgut/handlers/human_survey.py
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-2> (38)
   - *M* amgut/lib/data_access/ag_data_access.py
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-3> (27)
   - *M* amgut/lib/survey_supp.py
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-4> (7)
   - *M* amgut/lib/util.py
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-5> (34)
   - *M* amgut/static/js/american_gut.js
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-6> (15)
   - *M* amgut/templates/animal_survey.html
   <https://github.com/biocore/american-gut-web/pull/432/files#diff-7>
   (294)

Patch Links:

   - https://github.com/biocore/american-gut-web/pull/432.patch
   - https://github.com/biocore/american-gut-web/pull/432.diff

—
<https://github.com/biocore/american-gut-web/pull/432>.

wasade:
Thanks, @squirrelo! Reviewing statically now, will pull down at test locally tomorrow

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,229 @@
+--Create pet survey in database
+----------------------------------------------------------
+-- survey_group
+----------------------------------------------------------
+INSERT INTO survey_group (group_order, american, british) VALUES (-2, 'Pet Information', 'Pet Information');
```
wasade: why -2 here?

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,229 @@
...
+
+----------------------------------------------------------
+-- surveys
+----------------------------------------------------------
+INSERT INTO surveys (survey_id, survey_group) VALUES (2, -2);
+
+----------------------------------------------------------
+-- survey_question
+----------------------------------------------------------
+INSERT INTO survey_question (survey_question_id, question_shortname, american, british) VALUES (127, 'NAME', 'Name', 'Name');
```
wasade: Just a sanity check early, do any of the shortnames here conflict with any of the existing ones?

wasade: ...actually, if there isn't a `UNIQUE` constraint already then we may want to add one for shortname

amgut/templates/animal_survey.html
```diff
@@ -1,241 +1,77 @@
 {% extends sitebase.html %}
+{% from wtforms import validators %}
+{% block head %}
+<import >
+<script type="text/javascript">
+    function clone_pet() {
```
wasade: ?

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,229 @@
+--Create pet survey in database
```
wasade: this patch number conflicts with #428 i believe, but just making a note as its of course dependent on merge order

wasade:
Few minor comments, but otherwise good. Will poke and prod live tomorrow

amgut/db/patches/0009.sql
squirrelo: Yup, already noted. Luckly it's a simple fix since there is no db changes, just insertions.

amgut/db/patches/0009.sql
squirrelo: Arbitrarily chosen. Needs to be one that human surveys doesn't use, and human starts at -1, so I put pets at -2.

amgut/db/patches/0009.sql
squirrelo: added unique and it didn't break anything, so they are not in conflict.

amgut/templates/animal_survey.html
squirrelo: needed so you can add multiple other pets the pet in question lives with. Allows you to create a new dropdown if there is more than one.

squirrelo:
change pushed, ready for testing.

wasade:
Just FYI, I have not been able to verify this locally still. @squirrelo, I sent some chat messages about the issues. Will need to wait until Monday to resolve.

wasade:
Few items from a first interactive pass:

- submit button is very small and out of the way (see screenshot)
screen shot 2015-07-27 at 2 19 31 pm
- I was able to enter weird things like non-numeric characters for age
- Editing is not possible

amgut/db/patches/0009.sql
```diff
@@ -0,0 +1,230 @@
...
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (142, 'Never', 3);
+
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (143, 'Unspecified', 0);
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (143, 'High', 1);
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (143, 'Moderate', 2);
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (143, 'Low', 3);
+INSERT INTO survey_question_response (survey_question_id, response, display_index) VALUES (143, 'Never', 4);
+----------------------------------------------------------
+-- survey_question_triggers
+----------------------------------------------------------
```
adamrp: This comment (lines 228 through 230) can be removed, i think

amgut/lib/data_access/ag_data_access.py
```diff
@@ -625,12 +628,18 @@ def AGGetBarcodeMetadataAnimal(self, barcode):
         return return_res
 
     def getAnimalParticipants(self, ag_login_id):
-        results = self._sql.execute_proc_return_cursor(
-            'ag_get_animal_participants', [ag_login_id])
-
-        return_res = [row[0] for row in results]
-        results.close()
-        return return_res
+        sql = """SELECT participant_name FROM ag_animal_survey
```
adamrp: is `ag_animal_survey` the table from the old database where we were storing responses? Why do you need to get the participant name from more than one place? It seems like you should only need to get it from one place (and if you're currently having to look in multiple places, they should be consolidated into a single place -- after all, the questions on the pet survey didn't actually change, right?)

squirrelo: That is from the old table, yes. Do you want me to migrate everything over as part of this patch?

adamrp: Yeah, I think that makes sense to do. @wasade, thoughts? The main reason I'm concerned is that the function `getAnimalParticipants` might be used in places where it is assumed that a set of operations can be applied equally to each item in the output, which might not be the case if there is more than one source for the data.

wasade: It's only used in `get_menu_items`, and the animal survey handler, so as
long as it works with both of those were good? Recommend an appropriate doc
string though

squirrelo: If we want editing capability we need to migrate everything, otherwise we're good with it as is.

squirrelo: Thoughts on needing editing?

adamrp: The migration should be very simple, right? There are no new/different columns or anything so "old" surveys and "new" surveys have the same set of columns. So I don't think the migration should be a blocker for allowing editing. What else would be needed to permit pet survey editing? If there is little/nothing else, I think we should do it now.

squirrelo: Shoud be a simple mapping of answers, but won't be 100% sure until I test it. I'll update this when I have more info.

wasade: Thanks! It's one of those things that we need to do, and I worry it'll only increase in difficulty if delayed

squirrelo:
This should be ready for another review. Will have a gist ready for the migration soon.

wasade:
Thanks, @squirrelo! This is getting very close. Few comments:

Age/gender of associated humans is odd, see attached screenshot

Toilet water access is a bit confusing, but if that's how it was worded previously then better to retain

When selecting "other" for animal type, it should allow the user to enter the animal type.

I was able to enter non numeric values for the age

Could the "Age" question be rephrased as "Age (years)"?

screen shot 2015-07-30 at 3 51 22 pm

squirrelo:
Did you do a test database wipe and update? the picture looks like you have an older version of the survey patch in there. I can add the triggered question for pet type other and age integer check/renaming.

squirrelo:
Changes ready, needs a wipe/reload of the test database.

wasade:
Still able to specify non-numeric values for ages on both the pet and humans living with the pet

squirrelo:
rearranged the survey a bit to allow editing, and added editing/deleting capability for pets. Ready for another round of review.

wasade:
Thanks @squirrelo!!!

PR #433, opened 2015-07-24
Break out secondary survey links for better readability

squirrelo:
See screenshot. Hopefully more people will take the secondary surveys if the links are more easily picked up on.

screen shot 2015-07-24 at 2 27 48 pm


JWDebelius:
Will participants know that the title is a hyperlink? Should there be a second hyper link on the page. Like, I dunno "Go to the survey" at the end of the paragraph?

Otherwise, it looks good!

adamrp:
My personal opinion is that users who use the internet are accustomed to seeing blue and/or underlined text and thinking hyperlink, so it's probably fine as it is. So this has my :+1: after @squirrelo responds.

squirrelo:
I'd like to trust people can figure out the link is a link.

Issue #434, opened 2015-07-27
Add Travis CI


adamrp:
I think this should be a top priority.

squirrelo:
Done in #487

PR #430, opened 2015-07-21
Barcodes in own schema

squirrelo:
Depends on #429

This moves the barcode table and associated tables to their own schema, keeping the foreign keys from the ag schema.

PR #436, opened 2015-07-27
Break function out for database resets in tests

squirrelo:
~~Requires #430~~

This moves most of the functions that previously resided in the `ag` script to the `env_management.py` file. This allows changes added to the test checker decorator that reset the database after each test. 

This is a major push for #434 (adding travis testing)


adamrp:
@squirrelo this seems to contain commits that were merged in other recent PRs (e.g., #429, #430). Is there a reason for the duplication? Maybe you can pull upstream master to get around it?

squirrelo:
Yeah sorry, juggling many pull requests. Done.

adamrp:
No problem! Thanks :smile:

amgut/lib/data_access/env_management.py
```diff
@@ -0,0 +1,387 @@
...
+
+    # Get the cursor
+    cur = conn.cursor()
+
+    cur.execute('SET SEARCH_PATH TO ag, barcodes, public')
+
+    db_exists = _check_db_exists(AMGUT_CONFIG.database, cur)
+
+    # Check that the database does not already exist
+    if db_exists and force:
```
adamrp: I find this construction just slightly confusing, and would find a nested `if` clearer:
```python
if db_exists:
    if force:
        return
    raise ValueError("Database '{}' already present on the system".format(AMGUT_CONFIG.database))
```

adamrp:
Besides minor comment on clarity (not blocking), :+1:

PR #446, opened 2015-08-12
Delete test barcodes

squirrelo:
Adds patch to delete the large test barcodes from the system. Also allows verbose patching.

### DIFF
--- a/amgut/db/patches/0013.sql
+++ b/amgut/db/patches/0013.sql
@@ -47,6 +47,16 @@ LOOP
 		--Do nothing on exception because that means there are still barcodes attached to the kit
 	END;
 END LOOP;
+
+-- Wipe out test barcodes that are not attached to AG
+FOR bc IN
+	SELECT barcode FROM barcodes.barcode WHERE barcode::integer >= 800000000
+LOOP
+	DELETE FROM barcodes.barcode_exceptions WHERE barcode = bc;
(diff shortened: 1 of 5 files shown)


squirrelo:
Ready for review.

amgut/db/patches/0013.sql
```diff
@@ -0,0 +1,65 @@
...
+-- Remove the test barcodes (900000000 and above) so new barcode system works
+DO $do$
+DECLARE
+	bc varchar;
+	sid varchar;
+	agl uuid;
+	agk uuid;
+	pn varchar;
+BEGIN
+-- Wipe out test barcodes taht are already attached to kits and surveys
```
wasade: taht -> that

scripts/ag
```diff
@@ -66,6 +66,7 @@ def make(db, force):
     make_settings_table()
     click.echo("Applying patches to database")
     patch_db(verbose=True)
+    migrate_procedures('public', 'ag', verbose=True)
```
wasade: this call always excepts when i run it, does that happen in your env too?

squirrelo: it only excepts for me if you run it outside of make (aka run it on a database that already had it run). That's why I moved it here.

squirrelo:
Ready to go.

PR #455, opened 2015-08-18
Move around a bunch of barcodes

squirrelo:
This attaches barcodes to the proper projects.

### DIFF
--- a/amgut/db/patches/0017.sql
+++ b/amgut/db/patches/0017.sql
@@ -1,0 +1,195 @@
+-- August 17, 2015
+-- Rearrange barcodes to be attached to the correct projects
+Do $do$
+DECLARE
+    ag_pid bigint;
+    pid bigint;
+BEGIN
+ag_pid := (SELECT project_id FROM barcodes.project WHERE project = 'American Gut Project');
+
+-- Assign british gut kits to british gut and american gut
+CREATE TEMP TABLE hold_table (
+    barcode varchar,
+    project_id bigint
+);
+
+pid := (SELECT project_id FROM barcodes.project WHERE project = 'British Gut Project');
+
+INSERT INTO hold_table (project_id, barcode)
+    SELECT pid, barcode FROM (
+        SELECT barcode FROM ag.ag_kit_barcodes
+        JOIN ag.ag_kit USING (ag_kit_id)
+        WHERE lower(supplied_kit_id) LIKE '%bg\_%'
+        UNION
+        SELECT barcode from ag.ag_handout_barcodes
+        JOIN ag.ag_handout_kits USING (kit_id)
+        WHERE lower(kit_id) LIKE '%bg\_%') AS b;
+
+INSERT INTO hold_table (project_id, barcode)
+    SELECT ag_pid, barcode FROM hold_table;
+
+INSERT INTO barcodes.project_barcode (project_id, barcode)
+    SELECT project_id, barcode FROM hold_table
+    WHERE (project_id, barcode) NOT IN
+        (SELECT project_id, barcode FROM barcodes.project_barcode);
+
+DROP TABLE hold_table;
+
+-- rename Anxiety/Depression cohort to 'Mind and microbiome'
+UPDATE barcodes.project SET project = 'Mind and microbiome'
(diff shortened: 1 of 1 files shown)


amgut/db/patches/0017.sql
```diff
@@ -0,0 +1,195 @@
...
+BEGIN
+ag_pid := (SELECT project_id FROM barcodes.project WHERE project = 'American Gut Project');
+
+-- Assign british gut kits to british gut and american gut
+CREATE TEMP TABLE hold_table (
+    barcode varchar,
+    project_id bigint
+);
+
+pid := (SELECT project_id FROM barcodes.project WHERE project = 'British Gut Project');
```
wasade: should this be `bg_pid`?

squirrelo: All the projects are stored in the pid variable to cut down on namespace declarations.

squirrelo:
Looking it over, patch 16 should probably go at the end of this, and this should become patch 16.

amgut/db/patches/0017.sql
wasade: so why do `ag_pid`? what's the reason to be concerned about namespace declarations, aren't you more concerned about the potential namespace collisions?

wasade: ...not worth changing at this point though

squirrelo: `ag_pid` is constant over everything, since it is the american gut project id, so having it queried once and stored makes sense. `pid` is changing for each project that is being updated. Because you need to declare your variables at the time of function creation, it's easier to declare the two variables and reuse `pid` for each project.

amgut/db/patches/0017.sql
```diff
@@ -0,0 +1,195 @@
...
+    barcode varchar,
+    project_id bigint
+);
+
+INSERT INTO hold_table (project_id, barcode)
+    SELECT pid, barcode FROM (
+        SELECT barcode FROM ag.ag_kit_barcodes
+        JOIN ag.ag_kit USING (ag_kit_id)
+        WHERE lower(supplied_kit_id) LIKE '%asd\_%'
+        OR lower(supplied_kit_id) LIKE '%asdp\_%'
```
wasade: asd parent?

squirrelo: asd project

amgut/db/patches/0017.sql
```diff
@@ -0,0 +1,195 @@
...
+INSERT INTO project (project_id, project)
+    SELECT max(project_id)+1, 'Amnon Oral Timeseries' FROM project;
+
+pid := (SELECT project_id FROM barcodes.project WHERE project = 'Amnon Oral Timeseries');
+
+INSERT INTO hold_table (project_id, barcode)
+    SELECT pid, barcode FROM (
+        SELECT barcode FROM ag.ag_kit_barcodes
+        JOIN ag.ag_kit USING (ag_kit_id)
+        WHERE lower(supplied_kit_id) LIKE '%mts\_%'
```
wasade: three different prefixes?

squirrelo: Yup. No idea why, but that's how the kits were created.

amgut/db/patches/0017.sql
```diff
@@ -0,0 +1,195 @@
...
+        OR lower(supplied_kit_id) LIKE '%osc\_%'
+        OR lower(supplied_kit_id) LIKE '%pulse\_%'
+        UNION
+        SELECT barcode from ag.ag_handout_barcodes
+        JOIN ag.ag_handout_kits USING (kit_id)
+        WHERE lower(kit_id) LIKE '%mts\_%'
```
wasade: on all of these, the kit IDs are `lower`ed, why? Are the prefixes not stable in case?

squirrelo: Nope, some are uppercase, some are lowercase. Depends on who created the kit. I know at least for PGP there are both `pgp_` and `PGP_` kits.

wasade:
@squirrelo, i guess this is good to merge?

squirrelo:
Is to me

PR #459, opened 2015-08-19
Add tox.ini file for Flake8

squirrelo:


wasade:
why drive flake8 via tox instead of travis?

squirrelo:
because other linter plugins (such as for SublimeText) look for that file and can auto-set themselves using it.

squirrelo:
The other filename it looks for is `setup.cfg` so can move to that if wanted.

tox.ini
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 95
```
wasade: please leave this at 80

squirrelo: :( ok.

ElDeveloper: ping

squirrelo:
Change made

PR #465, opened 2015-08-24
Remove procedures

squirrelo:
This gets rid of all procedures and procedure calls in american gut codebase. The remaining procedures in use were single, simple calls that had no reason to be procedures at all.


amgut/lib/data_access/ag_data_access.py
```diff
@@ -380,40 +389,50 @@ def getAnimalParticipants(self, ag_login_id):
...
-        rows = results.fetchall()
-        col_names = self._get_col_names_from_cursor(results)
-        results.close()
+        sql = """SELECT  barcode, site_sampled, sample_date, sample_time,
+                    notes, status
+                 FROM ag_kit_barcodes akb
+                 INNER JOIN barcode USING (barcode)
+                 INNER JOIN ag_kit ak USING (ag_kit_id)
+                 WHERE (site_sampled IS NOT NULL AND site_sampled::text <> '')
+                 AND ag_login_id = %s AND participant_name = %s"""
```
wasade: this condition would evaluate faster if `ag_login_id` were first as that would immediately partition the set to a much smaller range of rows to interrogate further

amgut/lib/data_access/ag_data_access.py
```diff
@@ -380,40 +389,50 @@ def getAnimalParticipants(self, ag_login_id):
...
-
-        barcodes = [dict(zip(col_names, row)) for row in rows]
+        sql = """SELECT  barcode, site_sampled, sample_date, sample_time,
+                    notes, status
+                 FROM ag_kit_barcodes
+                 INNER JOIN barcode USING (barcode)
+                 INNER JOIN ag_kit USING(ag_kit_id)
+                 WHERE (environment_sampled IS NOT NULL AND
+                    environment_sampled::text <> '')
+                    AND ag_login_id = %s"""
```
wasade: same here re: ag_login_id

amgut/lib/data_access/ag_data_access.py
```diff
@@ -540,31 +564,33 @@ def ag_verify_kit_password_change_code(self, email, kitid, passcode):
...
+        kitid is the supplied_kit_id from the ag_kit table
         """
-        results = self._sql.execute_proc_return_cursor(
-            'ag_get_barcodes_by_kit', [kitID])
-        barcodes = [row[0] for row in results]
-        results.close()
-        return barcodes
+        sql = """SELECT barcode
+                 FROM ag_kit_barcodes
+                 INNER JOIN ag_kit kUSING(ag_kit_id)
```
wasade: bug that would bomb the query

this implies that this PR hasn't been exercised... can you verify locally please?

wasade:
...so the bulk of these procedures that were dropped weren't actually in use anywhere?

squirrelo:
Yes, ever since we moved to the tornado/postgres current system, every time anyone has edited the data access classes they removed calls to the procedure and added SQL instead. This is now the final changes to remove the last ones.

amgut/lib/data_access/ag_data_access.py
squirrelo: Where clause order doesn't matter in SQL, since the SQL parser will optimize the order itself as part of running the query. [See here](http://stackoverflow.com/questions/11436469/does-order-of-where-clauses-matter-in-sql).

wasade: sql engines are works of voodoo

amgut/lib/data_access/ag_data_access.py
```diff
@@ -529,9 +552,10 @@ def ag_update_kit_password(self, kit_id, password):
         """
         password = bcrypt.encrypt(password)
 
-        self.get_cursor().callproc('ag_update_kit_password',
-                                   [kit_id, password])
-        self.connection.commit()
+        sql = """UPDATE AG_KIT
```
ElDeveloper: Minor, but it seems like you've been writing the name of the tables with all lowercase, except here ...

ElDeveloper:
Code looks fine, but I'm hesitant to merge this without seeing that a test suite is executed or that the website is exercised fully (I do not have a working dev environment).

squirrelo:
We have no test suite for this, and adding one would take at least a week, so that is not feasible at this time. The whole point of these changes is to slim down the codebase so we can add tests for what we know is needed instead of the gigantic mess of code that previously was here.

ElDeveloper:
That's fine with me, have all of these been validate and exercised? I
just don't feel confident to merge something that I can't test.

squirrelo:
I’ve exercised them to the best of my ability. I'll also have my computer set up as a test server for a bit once everything is merged so we can run through the site with all the changes in place and see if there are any hiccups. I have a snapshot of the database for testing that I've been using.

josenavas:
Just throwing my 2 cents in here: probably good to set up a VM with a test environment, as we are going to do with Qiita?

wasade:
Absolutely

ElDeveloper:
OK, then I'm fine with this then.

PR #464, opened 2015-08-22
Select country on registration

squirrelo:
This makes the country list a selection instead of free text for registering a kit. This way at least we have controlled vocab for country of registration for later use in survey pulldown inferences.

~~Requires #462 first to avoid merge conflicts~~

PR #462, opened 2015-08-20
Alter dates times

squirrelo:
This makes all date and time columns use date and time format instead of varchar.

PR #466, opened 2015-08-24
Add country to zipcodes table

squirrelo:
Does what it says on the tin, wiping out anything that isn't obviously american or canadian in that table so it can repopulate properly with country.


squirrelo:
Ready for review, but merge #462 first so I can get merge conflicts resolved properly.

amgut/db/patches/0018.sql
```diff
@@ -0,0 +1,8 @@
+-- August 24, 2015
+-- add country column to zipcode table
+ALTER TABLE ag.zipcodes ADD COLUMN country varchar NOT NULL DEFAULT '';
+ALTER TABLE ag.zipcodes ALTER COLUMN country DROP DEFAULT;
+UPDATE ag.zipcodes SET country = 'United States' WHERE state IN ('AL','AK','AZ','AR','CA','CO','CT','DC','DE','FL','GA','HI','ID','IL','IN','IA','KS','KY','LA','ME','MD','MA','MI','MN','MS','MO','MT','NE','NV','NH','NJ','NM','NY','NC','ND','OH','OK','OR','PA','RI','SC','SD','TN','TX','UT','VT','VA','WA','WV','WI','WY');
```
wasade: GAZ is "United States of America"

amgut/db/patches/0018.sql
```diff
@@ -0,0 +1,8 @@
...
+UPDATE ag.zipcodes SET country = 'Canada' WHERE state IN ('AB','BC','MB','NB','NL','NS','NT','NU','ON','PE','QC','SK','YT');
+-- Delete the rest so the table can repopulate correctly
```
wasade: I dont understand this

squirrelo: The table will repopulate as things are pulled down, so any zipcodes that are british, mexican, ukranian, etc participants will be repopulated in the DB as proper geocode info instead of the mostly empty info we have for these zipcodes right now. They were added by hand since the table had no way of populating itself.

wasade: Will that set country details on the pulldown? If so, will the country detail be compatible with GAZ?

squirrelo: Countries added will be ISO-3166 as used by the rest of the information in AG, including the surveys, as well as Google in the geocoding.

amgut/db/patches/0018.sql
squirrelo: We need to figure country stuff out, since we have everything in the database as ISO-3166 and the `GAZ:United States of America` is wrong for the ontology.

squirrelo: OK, after talking with Gail, this is how the countries stuff shakes out:
1) All information stored in the database is in ISO 3166 format, so we can geocode from google with ease. 
2) On pulldown, a new column in the country table, `EBI`, will be used to convert the country in the DB to the country EBI expects.

wasade:
@squirrelo, conflicts but :+1: otherwise i think

squirrelo:
Conflict resolved. Also, this is now patch 0021 so it will not conflict with #464

PR #476, opened 2015-09-03
fix broken drop_test command

beaugunderson:
just noticed this was broken when I went to apply the new patches to my very old test database.


squirrelo:
:+1:

PR #477, opened 2015-09-04
Test framework

squirrelo:
This puts in test framework and an example SQL test for the data access class.


mortonjt:
Do you think it is possible to get a test database setup on this PR?

amgut/lib/data_access/tests/test_ag_data_access.py
```diff
@@ -0,0 +1,162 @@
...
+        # Verify password is not what we are changing it to
+        sql = 'SELECT kit_password FROM ag.ag_kit WHERE supplied_kit_id = %s'
+        old_pass = self.conn_handler.execute_fetchone(sql, [skid])[0]
+        self.assertFalse(bcrypt.verify('password', old_pass))
+
+        # Verify password change
+        test_sql = '''SELECT kit_password, pass_reset_code
+                      FROM ag.ag_kit
+                      WHERE supplied_kit_id = %s'''
+        try:
```
mortonjt: I'm just curious. What sort of exceptions are we trying to catch?  As a bystander, I wouldn't think having try excepts in a test case would be desirable ...

squirrelo: As it says, you need to reset the database to what it was no matter if the tests fail or not, hence the try/finally

squirrelo:
The timeframe for the test database is a bit out, since we need to have multiple people make sure it is scrubbed properly. This can at least be merged as a start, since it is working on test, which is part of any database dump. Then the next pull request will be the database and changes required to get that set up.

mortonjt:
I vote to get the test database merged with this PR.  Its the only way we can insure that the tests are consistent.

amgut/lib/data_access/tests/test_ag_data_access.py
mortonjt: I think this should go.  This is a sign of bad testing.  If the tests fail, then we'll always be able to re-fetch the test database from git

squirrelo: That's not how database testing works. If this test fails, and another test depends on the same data, the database is not consistent for the next test. Look at how qiita works: it actually drops the entire schema and recreates everything after each test function. We're avoiding that here for time sake.

mortonjt: I just looked at the qiita database test suite.

I haven't found a ___single___ try except statement in entire qiita database test suite.  In unittests, you want to raise exceptions, not catch them.  If you start adding try, except statements in your test cases, you are asking for poor test coverage.

Why can't we do the same thing that qiita does?  I vote to spend some proper time to thoroughly work through these test cases.  Rushing through them will seriously bite us in the butt later.  @wasade , @adamrp , @josenavas , @antgonza , @ElDeveloper , could you comment here to make sure that I'm not missing something.

josenavas: @squirrelo we had this discussion in the past in qiita. What are you trying to do here in the try/finally is what you have to put in your setUp/tearDown functions. Please, follow coding & testing good practices...

BTW the `conn_handler` and `ag_data`definition should go in the `setUp` as instance variables, not as class attributes as they're right now.

squirrelo: This is not a try/except, it's a try/finally. (Here is a link to how they work)[https://docs.python.org/2/tutorial/errors.html#defining-clean-up-actions]. Basically, this makes it so, whether the tested function succeeds or fails, the database is reset to how it was before the test started. This is important because if a test fails, you can have an inconsistent state in the database and won't know if the remaining tests failed because they actually failed, or because the database is in an inconsistent state and therefore the inputs to the tests were not as expected. 

There are no try/finally in qiita because, as I said, the schema is dropped and recreated after each test, meaning all the data in the database is reset regardless of what was being tested. We can't do that in this case because the test database takes about 30sec-1min to be created and patched up, which is prohibitive. That would be 30sec-1min of overhead per test function. When I was first starting out the O'Reilly postgres book really helpful.

josenavas:
My 2 cents - the test database should go in _before_ any tests goes in place. As happen with qiita, putting such database in place was a fair amount of work, however, it's value is incalculable at this point of the development effort. There is no point on writing test if those tests can be run only in a single machine.

We learned a lot from qiita, please do not incur in the same errors....

squirrelo:
@josenavas as stated before, this is a test FRAMEWORK not the actual tests, and the test id that is used in the example test will be there no matter what as it was in the database from the beginning. 

Please read the entire discussion before commenting.

amgut/lib/data_access/tests/test_ag_data_access.py
josenavas: @squirrelo we know how try/finally works... but that is not how testing works. PLEASE PLEASE PLEASE use the tearDown/setUp functions as that is why those function exists, do not reinvent the wheel...

josenavas:
@squirrelo please be nice - and make sure you are doing what you are saying. This [line\(https://github.com/biocore/american-gut-web/pull/477/files#diff-d1feb72b9489b758e7a6d0a58aa47f19R92) is adding a test.

Also, do not change definitions. If you're creating a FRAMEWORK that includes the DB to be able to write the tests.

rob-knight:
+1

On Sep 8, 2015, at 7:47 AM, Jose Navas <notifications@github.com> wrote:

@squirrelo <https://github.com/squirrelo> please be nice - and make sure
you are doing what you are saying. This [line(
https://github.com/biocore/american-gut-web/pull/477/files#diff-d1feb72b9489b758e7a6d0a58aa47f19R92)
is adding a test.

Also, do not change definitions. If you're creating a FRAMEWORK that
includes the DB to be able to write the tests.

—
<https://github.com/biocore/american-gut-web/pull/477#issuecomment-138393475>
.

amgut/lib/data_access/tests/test_ag_data_access.py
mortonjt: :+1:

wasade: Do the tests need to depend on specific database state? The production code
doesnt, so assuming state runs the risk of over fitting the tests. Most of
the tests do not need to modify the database, and those that do can do so
in a controlled fashion such that the complexity of reseting state can
likely be avoided entirely.

PR #478, opened 2015-09-10
Create test database

squirrelo:
This will create and load a test database using the gzipped full cleaned database.


mortonjt:
Nice!

Now, where will this gzipped file be located?  Will it be in this repo, or will it be elsewhere (i.e. [GFS](https://git-lfs.github.com/)).

If that test database will be in the repo, it'll need to be uploaded in this PR.

Also, is there a reason why it is gzipped?  Git doesn't do very well with detecting diffs on zip files.  And this will become unwieldly if we are making a lot of changes on this test database.

Do you think it is possible to unzip it and upload the .sql file?  The test database may not need to be that big.

squirrelo:
The DB will be in this PR as soon as it is approved. It's gzipped because the original DB is almost 40 megs, while zipped it's 4.7, so it saves a good deal of download time and repo space.

josenavas:
@squirrelo are you coordinating with @wasade there has been some ongoing discussion in potential re-visiting on how to do the tests, so it will be worth to communicate and don't replicate efforts

squirrelo:
Yes. This is part of the email that went out to all qiita-dev to look over the cleaned database.

mortonjt:
I strongly recommend trimming out this database.  We don't need to a monster database.  We just need enough tests to cover all of the critical test cases.

Looking at the [qiita](https://github.com/biocore/qiita/tree/master/qiita_db/support_files), it looks like all of the raw sql files are uploaded directly.  This will make it much more manageable to keep track of what changes are going into the database.

wasade:
I disagree. If we do not drop/reload on every test then the cost is small. The benefit is large as we've encountered frequent issues with unusual state in the db driven by it's history. 

We are _not_ uploading the real production database, so there should not be a need or concern to track changes in the contents of the test data. The test data can be housed in the repo or sourced (via travis) from ftp.microbio.me.

squirrelo:
Test database has been added. It's a 4.1MB file gzipped, so it is fine for size considering we have some 8+MB patches.

This is ready for review and testing.

mortonjt:
I really don't think having a gzipped file on github will buy us anything.  Can we store this in a separate location?

squirrelo:
It allows us to do a complete set up and test without having to rely on an outside source to still be available in the future.

amgut/lib/data_access/env_management.py
```diff
@@ -132,8 +133,12 @@ def make_settings_table():
 def populate_test_db():
     conn = SQLConnectionHandler()
 
-    with open(POPULATE_FP) as f:
-        conn.execute(f.read())
+    with gzip.open(POPULATE_FP, 'rb') as f:
+        test_db = f.read()
+
+    command = ['psql', '-d', AMGUT_CONFIG.database]
```
wasade: Should the code check and create the database if necessary?

amgut/lib/data_access/env_management.py
```diff
@@ -132,8 +133,12 @@ def make_settings_table():
...
+    proc = Popen(command, stdin=PIPE, stdout=PIPE)
+    proc.communicate(test_db)
```
wasade: Why `Popen` instead of `psycopg2`?

wasade: oh, I guess its because of the .sql files... okay

scripts/ag
```diff
@@ -61,16 +61,17 @@ def make(db, force):
...
     click.echo("Creating database")
     create_database(force)
-    build_and_initialize(verbose=True)
-    click.echo("Making settings table")
-    make_settings_table()
-    click.echo("Applying patches to database")
-    patch_db(verbose=True)
 
-    if db == 'test':
+    if db == 'production':
```
wasade: Could you add a comment to the "test" branch that build/initialize and setttings table creation aren't necessary as they are already included in the test database?

wasade:
:+1:, though please do not merge yet: I'd like to do one more pass over the test db

amgut/lib/data_access/env_management.py
squirrelo: `create_database()` should be called before this.

amgut/lib/data_access/env_management.py
squirrelo: Yeah, tried just psycopg2 but the way pd_dump gives files needs to be loaded with psql.

scripts/ag
squirrelo: Can do

amgut/lib/data_access/env_management.py
wasade: Okay, but should it error if called again?

```python
$ ./ag make test
/Users/mcdonadt/rs/american-gut-web/amgut/lib/config_manager.py:34: UserWarning: Extra main section option(s) found: 'debug'
  warnings.warn("Extra %s found: %r" % (set_type, extra))
Creating database
Traceback (most recent call last):
  File "./ag", line 99, in <module>
    cli()
  File "/Users/mcdonadt/miniconda3/envs/ag-web/lib/python2.7/site-packages/click/core.py", line 610, in __call__
    return self.main(*args, **kwargs)
  File "/Users/mcdonadt/miniconda3/envs/ag-web/lib/python2.7/site-packages/click/core.py", line 590, in main
    rv = self.invoke(ctx)
  File "/Users/mcdonadt/miniconda3/envs/ag-web/lib/python2.7/site-packages/click/core.py", line 936, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/mcdonadt/miniconda3/envs/ag-web/lib/python2.7/site-packages/click/core.py", line 782, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/mcdonadt/miniconda3/envs/ag-web/lib/python2.7/site-packages/click/core.py", line 416, in invoke
    return callback(*args, **kwargs)
  File "./ag", line 63, in make
    create_database(force)
  File "/Users/mcdonadt/rs/american-gut-web/amgut/lib/data_access/env_management.py", line 63, in create_database
    .format(AMGUT_CONFIG.database))
ValueError: Database 'ag2' already present on the system
```

squirrelo: Yes, although that should probably be a cleaner error. We can set up another thing like we have in qiita, `ag clean_test` to wipe out the DB and rebuild it.

josenavas:
I would like to review this too, however I will not have time until tonight.

wasade:
Creation of the database presently throws role errors. These can be removed with:

```bash
gunzip amgut/db/ag_test_patch22.sql
sed -i bak 's/jshorens/postgres/g' amgut/db/ag_test_patch22.sql
gzip amgut/db/ag_test_patch22.sql
```

(though, a fancy 1-liner is totally in reach there).

Test database has references to labadmin, can these be removed? These can be found if you just open the .sql in an editor. There are also references to "testingfoo" 

There are a few mentions to functions like `ag_authenticate_user`, haven't those been tossed?

Is the table `promoted_survey_ids` necessary?

The following tables appear to be empty, should they be removed?

- `zipcodes`

...

`plate` and `plate_barcode`

wasade:
Sorry, accidentally hit tab and space. Comment was premature

squirrelo:
Empty tables are emptied on purpose. The external survey answers/sources tables are needed to store external surveys once we have them, and zipcodes is for the zipcodes during geocoding by labadmin. `promoted_survey_ids` was also needed for labadmin. This is why there are some labadmin specific things in here, since it's all one database technically. There are also the procedures still in here since we haven't 100% removed all calls to them yet.

In light of some of these things, we will most likely need to do one more test database update to remove things once we have added tests and cleaned up the code completely. There are some things that are being relied on now that will not be once we clean things up.

squirrelo:
Changes made, ready for review.

wasade:
@josenavas, have you had a chance to review as well? Just pinging as you mentioned that you wanted to take a pass (and it would be awesome if you're able)

josenavas:
Thanks for the ping, going over it right now...

amgut/lib/data_access/env_management.py
josenavas: I've look a bit around. The reason what this is the case is because in the pg_dump, there are clauses like this one:
```
COPY survey_question_response_type (survey_question_id, survey_response_type) FROM stdin;
```
and then a list of values - which are given as if they where given form the standard input.

IMO, it will be better if we can use psycopg2, and you can do that by generating the pg_dump passing the following parameter:
```
--column-inserts             dump data as INSERT commands with column names
```

josenavas: And additionally, this one:
```
--inserts                    dump data as INSERT commands, rather than COPY
```

josenavas:
There is a lot of code in here that is a duplicate from qiita. However, in Qiita, the code has been improved - and it is not being improved in here. IMO, there is a lot of calls to `Popen` and AFAIK we are not doing any of those from Qiita. It might be useful to take a look in there and bring the knowledge on how to do that using the python libraries rather than continuous calls to `Popen`. This will give us more control on the execution.

josenavas:
BTW, regarding the test database. I could load it but I'm not familiar enough with the layout to know if there is any personal information that is being published in here.

amgut/lib/data_access/env_management.py
squirrelo: Looking at the docs, this makes it extremely, extremely slow to load data in, and should really only be used for moving to non-postgres databases. What do we gain by not using pg_dump, especially when we already do outside calls to it in other places (specifically dropdb)?

squirrelo:
There is currently a push for as little change as possible until we have tests in place. I would be more comfortable with doing the qiita-based upgrades after the tests are in and we can make sure functionality is not broken.

josenavas:
@wasade I give my point of view, is up to you.

squirrelo:
Any final word on this?

wasade:
@josenavas, qiita is populating a small test db whereas this is a full and encompassing db. The use is different. Calls to `Popen` are not by themselves bad, and fit well within one of the major paradigms of Python. I think the bigger question is: is this sufficient? Secondary question is: can we do it better? I think both of those are "yes", but I don't think the secondary point should not block merge.

PR #483, opened 2015-09-18
Migrate old unmappable questions

squirrelo:
This adds an easy way to retire questions. Retired questions will not show in the surveys, but will show in metadata pulldowns. 

This also moves over the old plants question as a retired question, allowing it to be pulled down.

### DIFF
--- a/amgut/db/patches/0023.sql
+++ b/amgut/db/patches/0023.sql
@@ -35,18 +35,47 @@ INSERT INTO survey_question_retired (survey_id, survey_question_id) VALUES (1, 1
 -- Promote answers
 DO $do$
 DECLARE
-    top varchar;
+    ans varchar;
     survey varchar;
     pid varchar;
     login varchar;
 BEGIN
     -- Make sure not in test database so there's actually something to promote
-    IF SELECT NOT EXISTS(SELECT 1
-                         FROM information_schema.tables 
-                         WHERE table_schema = 'ag' AND table_name = 'ag_human_survey')
+    IF NOT EXISTS(SELECT 1 FROM information_schema.tables WHERE table_schema = 'ag' AND table_name = 'ag_human_survey')
     THEN
         RETURN;
     END IF;
 
-    
+    FOR pid, login IN
+        SELECT participant_name, ag_login_id FROM ag.ag_human_survey
+    LOOP
+        SELECT survey_id FROM ag.ag_login_surveys WHERE participant_name = pid AND ag_login_id::varchar = login INTO survey;
+        -- If unpromoted survey, don't convert answers
+        IF survey = NULL
+        THEN
+            CONTINUE;
+        END IF;
+        -- Assign uninterpretable answers to 'Unspecified'
+        SELECT CASE(types_of_plants)
+           WHEN '28' THEN '21 to 30'
+           WHEN 'Less than 5' THEN 'Less than 5'
+           WHEN '6 to 10' THEN '6 to 10'
+           WHEN '11 to 20' THEN '11 to 20'
+           WHEN '21 to 30' THEN '21 to 30'
+           WHEN 'More than 30' THEN 'More than 30'
+           ELSE 'Unspecified'
+        END
+        FROM ag.ag_human_survey WHERE participant_name = pid AND ag_login_id::varchar = login INTO ans;
+
+        INSERT INTO ag.survey_answers (survey_id, survey_question_id, response) VALUES (survey, 146, ans);
+    END LOOP;
 END $do$
+
+-- Make the rest of the answers Unspecified
+INSERT INTO ag.survey_answers (survey_id, survey_question_id, response)
+    SELECT survey_id, 146, 'Unspecified'
+    FROM ag.ag_login_surveys
+    WHERE survey_id NOT IN (
+        SELECT survey_id
+        FROM ag.survey_answers
+        WHERE survey_question_id = 146);
--- a/amgut/db/patches/0023.sql
+++ b/amgut/db/patches/0023.sql
@@ -1,0 +1,13 @@
+-- Sept 16, 2015
+-- Add way to retire questions
+--promote unmappable questions from old survey
+
+CREATE TABLE ag.survey_question_retired (
+	survey_id            bigint  NOT NULL,
+	survey_question_id   bigint  NOT NULL,
+	CONSTRAINT pk_survey_question_retired PRIMARY KEY ( survey_id, survey_question_id )
+ );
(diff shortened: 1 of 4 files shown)


rob-knight:
If Josh could get a review on this ASAP we'd appreciate it. Thanks!

On Sep 17, 2015, at 6:10 PM, Joshua Shorenstein <notifications@github.com>
wrote:

This adds an easy way to retire questions. Retired questions will not show
in the surveys, but will show in metadata pulldowns.

This also moves over the old plants question as a retired question,
allowing it to be pulled down.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/biocore/american-gut-web/pull/483
Commit Summary

   - add retired_survey_question table
   - add plants question
   - migrate over data

File Changes

   - *M* amgut/db/ag.dbs
   <https://github.com/biocore/american-gut-web/pull/483/files#diff-0> (17)
   - *M* amgut/db/ag.html
   <https://github.com/biocore/american-gut-web/pull/483/files#diff-1>
   (194)
   - *A* amgut/db/patches/0023.sql
   <https://github.com/biocore/american-gut-web/pull/483/files#diff-2> (81)

Patch Links:

   - https://github.com/biocore/american-gut-web/pull/483.patch
   - https://github.com/biocore/american-gut-web/pull/483.diff

—
<https://github.com/biocore/american-gut-web/pull/483>.

wasade:
This doesn't remap the span of responses to the questions, does it?

squirrelo:
Nope, direct port over.

amgut/db/patches/0023.sql
```diff
@@ -0,0 +1,73 @@
...
+
+        INSERT INTO ag.survey_answers (survey_id, survey_question_id, response) VALUES (survey, 146, ans);
+    END LOOP;
+END $do$
+
+-- Make the rest of the answers Unspecified
+INSERT INTO ag.survey_answers (survey_id, survey_question_id, response)
+    SELECT survey_id, 146, 'Unspecified'
+    FROM ag.ag_login_surveys
+    WHERE survey_id NOT IN (
```
josenavas: If there is a NULL in the inner select here, this will not work correctly. Did you double check that the inner select is not returning any null?

josenavas:
My understanding was that all the questions from the old survey need to be promoted so they appear in the metadata pull down. I see 2 paths:

1. Promote all the questions in here
2. Do only the database changes in here. Then do another PR promoting all the questions.

I personally don't like PR that are doing partial fixes...

rob-knight:
We (Josh, Justine, Gail Embriette and I) reviewed the questions that have
not been promoted yet.

Essentially, plants, which will reappear soon, is worth promoting.

Other data calculated from diet are in our opinion not useful. We will
however preserve old data by making Justine's spreadsheet publicly
available from the AG web site and perhaps data dryad, including column
mapping old to new id (for people who have the old data). New data
calculated from vioscreen will be included going forward as summary. Full
per-item data will be made available but not in EBI.

Other missing questions, mostly disease-related, have either been promoted
already or are better handled with third-party survey going forward.

Focus is on doing clean new submission rather than preserving patchwork of
existing submissions/id styles. The mapping table will solve (at some
inconvenience) the problems for those who already have copies of downloaded
data.

Please let us know if any concerns but this will greatly simplify life
going forward while making it still possible to do analysis with old
metadata if absolutely essential.

Rob

On Sep 17, 2015, at 9:41 PM, Jose Navas <notifications@github.com> wrote:

My understanding was that all the questions from the old survey need to be
promoted so they appear in the metadata pull down. I see 2 paths:

   1. Promote all the questions in here
   2. Do only the database changes in here. Then do another PR promoting
   all the questions.

I personally don't like PR that are doing partial fixes...

—
<https://github.com/biocore/american-gut-web/pull/483#issuecomment-141342260>
.

josenavas:
Ok thanks @rob-knight 

If there is only such question to promote, then this PR is ok, besides the comment that I did in the code.

josenavas:
Sorry, just forgot one think.

I noticed that the questions is promoted, but I haven't seen any change in the functionality that will make that question to not appear (unless I'm missing something).

If a question will not appear on the survey based on the new column added here, shouldn't be changes in the retrieval of the survey questions that checks for that column being FALSE?

rob-knight:
Yes I agree it's surprising but the other cases are either already handled
(eg diabetes), will be handled better with aux surveys (migraines etc.), or
have been superseded by much better but superficially similar and therefore
confusing data (fiber, calories, protein etc.) so making the old metadata
publicly available via mapping files etc. rather than storing forever in
much more complex db structure than is otherwise necessary made most sense.

On Sep 17, 2015, at 9:53 PM, Jose Navas <notifications@github.com> wrote:

Ok thanks @rob-knight <https://github.com/rob-knight>

If there is only such question to promote, then this PR is ok, besides the
comment that I did in the code.

—
<https://github.com/biocore/american-gut-web/pull/483#issuecomment-141343857>
.

amgut/db/patches/0023.sql
squirrelo: There is no way for it to be null, as this is what associates the survey with the barcode and answers. It was also tested on a dump of the database.

squirrelo:
@josenavas The check is the changes in `survey.py` since that is what creates the survey. The added check for retired being false allows retired survey questions to not appear.

amgut/db/patches/0023.sql
```diff
@@ -0,0 +1,73 @@
+-- Sept 16, 2015
+-- Add way to retire questions
+--promote unmappable questions from old survey
```
wasade: minor, but we're only promoting a single question

amgut/db/patches/0023.sql
wasade: `survey_id` is the primary key of `survey_answers`, so `NULL` is not possible

amgut/lib/data_access/survey.py
```diff
@@ -194,18 +194,22 @@ class Group(object):
...
         qs = [Question.factory(x[0], n) for x in db_conn.execute_fetchall('''
-            select gq.survey_question_id
-            from {0} sg join {1} gq on sg.group_order = gq.survey_group
-            where sg.group_order = %s
-            order by gq.display_index
+            SELECT gq.survey_question_id
+            FROM {0} sg
+            JOIN {1} gq ON sg.group_order = gq.survey_group
+            LEFT JOIN {2} sq USING (survey_question_id)
+            WHERE sg.group_order = %s AND sq.retired = FALSE
```
wasade: I like this a lot as it naturally supports future retired questions if they arise.

wasade:
This looks good, and I'm glad the consensus was that only a single question needed to be promoted.

PR #492, opened 2015-09-24
Rename columns

squirrelo:
After going over the pulldown with @ackermag we found a couple column headers that weren't quite clear. This renames them.


wasade:
@JWDebelius, we need to update the data dictionary, right?

JWDebelius:
Probably. The PR is still open.

PR #495, opened 2015-09-27
add setup and test folder to setup.py

squirrelo:


amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -3,6 +3,12 @@
 
 
 class TestAGDataAccess(TestCase):
+    def setUp(self):
+        ag_data = AGDataAccess()
```
wasade: needs a `self.ag_data =` ..

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -3,6 +3,12 @@
...
+
+    def tearDown(self):
+        del ag_data
```
wasade: needs `self`

wasade: why is this explicit?

PR #500, opened 2015-09-27
TST: Adding in test cases for getConsent, getParticipantSamples, getE…

mortonjt:
…nvironmentalSamples, getAvailableBarcodes

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -286,7 +286,8 @@ class AGDataAccess(object):
                                 ag_login_surveys agl
                                 USING (ag_login_id, participant_name)
                            WHERE agl.survey_id=%s""", [survey_id])
-            colnames = [x[0] for x in cur.description] # Can refactor this with _get_col_names_from_cursor()
+            # Can refactor this with _get_col_names_from_cursor()
+            colnames = [x[0] for x in cur.description]
             result = cur.fetchone()
             if result:
                 result = {k: v for k, v in zip(colnames, result)}
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -269,6 +269,7 @@ class AGDataAccess(object):
             curr.execute(sql, [ag_login_id, participant_name])
 
     def getConsent(self, survey_id):
+        conn_handler = SQL
         conn_handler = SQLConnectionHandler()
         with conn_handler.get_postgres_cursor() as cur:
             cur.execute("""SELECT agc.participant_name,
@@ -286,7 +287,7 @@ class AGDataAccess(object):
                                 ag_login_surveys agl
                                 USING (ag_login_id, participant_name)
                            WHERE agl.survey_id=%s""", [survey_id])
-            colnames = [x[0] for x in cur.description]
+            colnames = [x[0] for x in cur.description] # Can refactor this with _get_col_names_from_cursor()
             result = cur.fetchone()
             if result:
                 result = {k: v for k, v in zip(colnames, result)}
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -269,7 +269,6 @@ class AGDataAccess(object):
             curr.execute(sql, [ag_login_id, participant_name])
 
     def getConsent(self, survey_id):
-        conn_handler = SQL
         conn_handler = SQLConnectionHandler()
         with conn_handler.get_postgres_cursor() as cur:
             cur.execute("""SELECT agc.participant_name,
--- a/amgut/lib/data_access/test/test_ag_data_access.py
+++ b/amgut/lib/data_access/test/test_ag_data_access.py
@@ -1,6 +1,6 @@
 from unittest import TestCase, main
 from amgut.lib.data_access.ag_data_access import AGDataAccess
-
+import datetime
 
 class TestAGDataAccess(TestCase):
     def setUp(self):
@@ -8,21 +8,6 @@ class TestAGDataAccess(TestCase):
 
     def tearDown(self):
         del self.ag_data
-
-    def test___init__(self):
-        raise NotImplementedError()
-
-    def test___del__(self):
-        raise NotImplementedError()
-
-    def test_get_cursor(self):
-        raise NotImplementedError()
-
-    def test__open_connection(self):
-        raise NotImplementedError()
-
-    def test__get_col_names_from_cursor(self):
-        raise NotImplementedError()
 
     def test_authenticateWebAppUser(self):
         raise NotImplementedError()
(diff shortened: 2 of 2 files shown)


squirrelo:
flake8 errors

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -43,7 +28,20 @@ def test_deleteAGParticipantSurvey(self):
         raise NotImplementedError()
 
     def test_getConsent(self):
```
squirrelo: Also need to test when survey_id does not exist on database.

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -64,13 +62,127 @@ def test_getAnimalParticipants(self):
         raise NotImplementedError()
 
     def test_getParticipantSamples(self):
-        raise NotImplementedError()
+        res = self.ag_data.getParticipantSamples("d8592c74-9694-2135-e040-8a80115d6401", "REMOVED")
```
squirrelo: Also need to test where samples are over multiple kits/in a single kit.

squirrelo: Alos when ID does not exist in DB

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -64,13 +62,127 @@ def test_getAnimalParticipants(self):
...
-        raise NotImplementedError()
+        res = self.ag_data.getEnvironmentalSamples("d6b0f287-b9d9-40d4-82fd-a8fd3db6c476")
+        exp = [{'status': None, 'sample_time': datetime.time(21, 45),
+                'notes': 'REMOVED', 'barcode': '000028433',
+                'sample_date': datetime.date(2015, 6, 7), 'site_sampled': None}]
+        self.assertItemsEqual(res, exp)
 
     def test_getAvailableBarcodes(self):
-        raise NotImplementedError()
+        res = self.ag_data.getAvailableBarcodes("d8592c74-9694-2135-e040-8a80115d6401")
```
squirrelo: Need test for when login_id not in DB, and over multiple kits/single kit as two tests.

squirrelo:
Few comments

amgut/lib/data_access/ag_data_access.py
```diff
@@ -286,7 +286,7 @@ def getConsent(self, survey_id):
                                 ag_login_surveys agl
                                 USING (ag_login_id, participant_name)
                            WHERE agl.survey_id=%s""", [survey_id])
-            colnames = [x[0] for x in cur.description]
+            colnames = [x[0] for x in cur.description] # Can refactor this with _get_col_names_from_cursor()
```
wasade: this looks gt 80 chars, flake8 is probably flaking out right now

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -43,7 +29,20 @@ def test_deleteAGParticipantSurvey(self):
...
+
+        res = self.ag_data.getConsent("8b2b45bb3390b585")
+        exp = {'date_signed': 'None',
+               'assent_obtainer': None,
+               'age_range': None,
+               'parent_1_name': 'REMOVED',
+               'participant_email': 'REMOVED',
+               'parent_2_name': 'REMOVED',
+               'ag_login_id': '000fc4cd-8fa4-db8b-e050-8a800c5d02b5',
+               'deceased_parent': 'false',
```
wasade: can you create an issue on this? given that `is_juvenile` is actually boolean, this probably should be too

mortonjt:
Addressed @squirrelo @wasade comments

amgut/lib/data_access/test/test_ag_data_access.py
mortonjt: Created!

amgut/lib/data_access/test/test_ag_data_access.py
mortonjt: Done

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -64,13 +67,162 @@ def test_getAnimalParticipants(self):
...
+        self.assertItemsEqual(res, exp)
+
+        i = "d6b0f287-b9d9-40d4-82fd-a8fd3db6c476"
+        res = self.ag_data.getAvailableBarcodes(i)
+        exp = ['000028434']
+        self.assertItemsEqual(res, exp)
+
+    def test_getAvailableBarcodesNotPresent(self):
+        i = '00000000-0000-0000-0000-000000000000'
+        res = self.ag_data.getAvailableBarcodes(i)
```
wasade: should this error since it isn't a valid id?

wasade: (totally open question)

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -97,7 +249,13 @@ def test_ag_verify_kit_password_change_code(self):
         raise NotImplementedError()
 
     def test_getBarcodesByKit(self):
-        raise NotImplementedError()
+        res = self.ag_data.getBarcodesByKit('tst_iqyYo')
+        exp = ['000027561']
+        self.assertItemsEqual(res, exp)
+
+    def test_getBarcodesByKitNotPresent(self):
+        res = self.ag_data.getBarcodesByKit('42')
```
wasade: same here

amgut/lib/data_access/test/test_ag_data_access.py
mortonjt: I wasn't sure what exactly isn't a valid id.

It definitely didn't error - it just wasn't in the database

wasade: not valid as in not in the db

mortonjt: If it isn't in the db, what would be an appropriate way to handle such a query?

Right now, it is either returning `None` or `[]`

wasade: Should it raise?

squirrelo: Well, there are two ways we can go about it: error here and catch it in the tornado interface, or return nothing here and just act as if it's blank. I'm inclined to say error just because we can tehn see what went wrong and how, and fix it.

squirrelo: Having said that, for tests right now on this type of functionality, probably best to leave it not erroring since we have no idea what the errors will do to tornado.

ElDeveloper: I think I prefer to raise an error instead of returning [] or None or {}.

mortonjt: Seems that we are a leaning towards raising errors, but are a little conflicted.

Should I go head and just incorporate `raise ValueErrors` then?

wasade: Perhaps subclass exception (at the module level) and raise MissingID?

squirrelo: Yes, agreed, but I'd say for now open an issue as we have no idea what that will do to the tornado interface.

wasade: oh. excellent point. :/

mortonjt: What do you think about instead of raising an exception, raise a warning.  That way, we can still debug without potentially disturbing the tornado interface?

squirrelo: warning will still kill the interface.

squirrelo: wait, my bad, no they don't. I'd say just test it as is for now and open an issue. It will be quick to fix once we have tests in for tornado.

mortonjt:
Anything else that needs to be done before I commence squashing?

wasade:
i dont think so. i don't think there is a need to squash commits unless
you'd like to

PR #503, opened 2015-09-27
ENH: Improve checkPrintResults and tables

ElDeveloper:
In trying to improve the state of things with `DataAccess.checkPrintResults`,
this PR modifies the `ag_handout_kits` column to be a boolean instead of a
varchar type (patch 0026.sql).

By changing this to return a bool type in Python, we were able to remove an
`if` statement that accounted for cases where the `kit_id` is not present in
the table.

:ship:


squirrelo:
`test_checkPrintResults` has an assertion error

ElDeveloper:
Thanks @squirrelo, just fixed that!

amgut/lib/data_access/ag_data_access.py
```diff
@@ -580,12 +578,29 @@ def getBarcodesByKit(self, kitid):
...
+        results would be available for a non-existent `kit_id`.
+        """
         sql = "SELECT print_results FROM ag_handout_kits WHERE kit_id = %s"
         results = self._sql.execute_fetchone(sql, [kit_id])
         if results is None:
-            return None
+            return False
         else:
-            return results[0].strip()
+            return results[0]
```
mortonjt: Now, is this guaranteed to return a bool?
From the old line of code, it appeared that `results[0]` was a string - you can't `strip()` bools.
Would it be worthwhile to case?

ElDeveloper: It is guaranteed to be a Python bool, because the column is set to be boolean and psycopg deals wit this nicely so no need to preprocess the output before coercing it into a recognizable type.

mortonjt: Awesome!

mortonjt:
I can't see any immediate problems with this

PR #515, opened 2015-09-29
More tests

squirrelo:
This adds tests for handoutCheck, check_access, verifyKit

### DIFF
--- a/amgut/lib/data_access/test/test_ag_data_access.py
+++ b/amgut/lib/data_access/test/test_ag_data_access.py
@@ -423,7 +423,24 @@ class TestAGDataAccess(TestCase):
         self.assertEqual(res, [])
 
     def test_verifyKit(self):
-        raise NotImplementedError()
+        # Test verifying works
+        kit = self.ag_data._get_unverified_kits()[0]
+        obs = self.ag_data.getAGKitDetails('tst_ODmhG')
+        self.assertEqual(obs['kit_verified'], 'y')
+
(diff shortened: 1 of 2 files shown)


amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -426,19 +423,51 @@ def test_getAvailableBarcodesNotPresent(self):
...
+        kit = self.ag_data._get_unverified_kits()[0]
+        self.ag_data.verifyKit(kit)
+        obs = self.ag_data.getAGKitDetails(kit)
+        self.assertEqual(obs['kit_verified'], 'y')
 
-    def test_getMapMarkers(self):
-        raise NotImplementedError()
+        # Test verifying a non-existant kit
+        # TODO: make raise an error
+        self.ag_data.getAGKitDetails('NOTAREALKITID')
```
mortonjt: Why is this here?  Even if it doesn't raise an error, should there be some sort of check?

squirrelo: It will eventually raise an error, so it's here to make sure the test is updated once it is updated. This will make the test fail once the update is in place, so it's not forgotten.

mortonjt: Sounds good

mortonjt:
I'm not seeing any outstanding problems in this PR.

PR #516, opened 2015-09-29
Change barcode for testing

squirrelo:
This makes logParticipantSample test work. Previously it was using a barcode used in another test, so the interaction between the two was causing stochastic failures depending on the order the tests were run.


squirrelo:
This is ready to go.

PR #519, opened 2015-10-01
Add transaction object

squirrelo:
This transitions away from the hodge-podge of SQL handling and standardizes on the transaction object from qiita.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -13,6 +13,7 @@ Centralized database access for the American Gut web portal
 """
 
 import logging
+from uuid import UUID
 
 import psycopg2
 from passlib.hash import bcrypt
@@ -83,16 +84,16 @@ class AGDataAccess(object):
                 WHERE supplied_kit_id = %s"""
             TRN.add(sql, [username])
             row = TRN.execute_fetchindex()
-            if row:
-                results = dict(row[0])
-
-                if not bcrypt.verify(password, results['kit_password']):
-                    return False
-                results['ag_login_id'] = str(results['ag_login_id'])
-
-                return results
-            else:
+            if not row:
                 return False
+
+            results = dict(row[0])
+
+            if not bcrypt.verify(password, results['kit_password']):
+                return False
+            results['ag_login_id'] = str(results['ag_login_id'])
+
+            return results
 
     def check_login_exists(self, email):
         """Checks if email for login already exists on system
@@ -112,9 +113,9 @@ class AGDataAccess(object):
             sql = "SELECT ag_login_id FROM ag_login WHERE LOWER(email) = %s"
             TRN.add(sql, [clean_email])
             value = TRN.execute_fetchindex()
-            if not value:
-                return None
-            return value[0][0]
+            if value:
+                value = value[0][0]
+            return value
 
     def addAGLogin(self, email, name, address, city, state, zip_, country):
         """Adds a new login or returns the login_id if email already exists
@@ -156,6 +157,23 @@ class AGDataAccess(object):
             return ag_login_id
 
     def getAGBarcodeDetails(self, barcode):
+        """Returns information about the barcode from both AG and standard info
+
+        Parameters
+        ----------
+        barcode : str
+            Barcode to get information for
+
+        Returns
+        -------
+        dict
+            All barcode info, keyed to column name
+
+        Raises
+        ------
+        ValueError
+            Barcode not found in AG information tables
+        """
         sql = """SELECT  email,
                     cast(ag_kit_barcode_id as varchar(100)),
                     cast(ag_kit_id as varchar(100)),
@@ -192,26 +210,38 @@ class AGDataAccess(object):
             return dict(row[0])
 
     def registerHandoutKit(self, ag_login_id, supplied_kit_id):
-        """
+        """Registeres a handout kit to a user
+
+        Parameters
+        ----------
+        ag_login_id : str
+            UUID4 formatted string of login ID to associate with kit
+        supplied_kit_id : str
+            kit ID for the handout kit
+
         Returns
         -------
         bool
             True:  success
             False: insert failed due to IntegrityError
+
+        Raises
+        ------
+        ValueError
+            Non-UUID4 value sent as ag_login_id
         """
         with TRN:
+            uuid_check = UUID(ag_login_id)
+            if not uuid_check.variant == 4:
+                raise ValueError("Non-UUID4 value passed: %s" % ag_login_id)
+
             printresults = self.checkPrintResults(supplied_kit_id)
-            if printresults is False:
-                printresults = 'n'
             # make sure login_id and skid exists
             sql = """SELECT EXISTS(SELECT *
                                    FROM ag.ag_login
                                    WHERE ag_login_id = %s)"""
             TRN.add(sql, [ag_login_id])
-            try:
-                exists = TRN.execute_fetchlast()
-            except ValueError:
-                raise ValueError("ag_login_id is not a UUID: %s" % ag_login_id)
+            exists = TRN.execute_fetchlast()
             if not exists:
                 return False
             sql = """SELECT EXISTS(SELECT *
@@ -309,7 +339,6 @@ class AGDataAccess(object):
                            JOIN ag_login_surveys agl
                            USING (ag_login_id, participant_name)
                            WHERE agl.survey_id = %s""", [survey_id])
-            # Can refactor this with _get_col_names_from_cursor()
             result = TRN.execute_fetchindex()
             if not result:
                 raise ValueError("Survey ID does not exist in DB: %s" %
@@ -707,6 +736,11 @@ class AGDataAccess(object):
         list of dict
             A list of registration information associated with a common login
             ID.
+
+        Raises
+        ------
+        ValueError
+            Unknown ag_login_id passed
         """
         with TRN:
             sql = """SELECT ag_login_id, email, name, address, city, state,
@@ -734,6 +768,11 @@ class AGDataAccess(object):
         -------
         str or None
             The survey ID, or None if a survey ID cannot be found.
+
+        Raises
+        ------
+        ValueError
+            Unknown ag_login_id or participant_name passed
         """
         with TRN:
             sql = """SELECT survey_id
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -17,8 +17,7 @@ import logging
 import psycopg2
 from passlib.hash import bcrypt
 
-from amgut.lib.data_access.sql_connection import SQLConnectionHandler
-from amgut.lib.config_manager import AMGUT_CONFIG
+from amgut.lib.data_access.sql_connection import TRN
 
 
 # character sets for kit id, passwords and verification codes
@@ -64,42 +63,6 @@ class AGDataAccess(object):
                      'Sole of shoe',
                      'Water']
 
-    def __init__(self, con=None):
-        self.connection = None
-        if con is None:
-            self._open_connection()
-        else:
-            self.connection = con
-        cur = self.get_cursor()
-        cur.execute('set search_path TO ag, barcodes, public')
-
-        self._sql = SQLConnectionHandler(con)
-
-    def __del__(self):
-        self.connection.close()
-
-    def get_cursor(self):
-        if self.connection.closed:
-            self._open_connection()
-
-        return self.connection.cursor()
-
-    def _open_connection(self):
-        self.connection = psycopg2.connect(
-            user=AMGUT_CONFIG.user, password=AMGUT_CONFIG.password,
-            database=AMGUT_CONFIG.database, host=AMGUT_CONFIG.host,
-            port=AMGUT_CONFIG.port)
-
-    #####################################
-    # Helper Functions
-    #####################################
-
-    def _get_col_names_from_cursor(self, cur):
-        if cur.description:
-            return [x[0] for x in cur.description]
-        else:
-            return []
-
     #####################################
     # Users
     #####################################
@@ -111,23 +74,25 @@ class AGDataAccess(object):
         web_app_user table. If successful, a dict with user innformation is
         returned. If not, the function returns False.
         """
-        sql = """SELECT  cast(ag_login_id as varchar(100)) as ag_login_id,
-                  email, name, address, city,
-                  state, zip, country,kit_password
-                FROM ag_login
-            INNER JOIN ag_kit USING (ag_login_id)
-            WHERE supplied_kit_id = %s"""
-        row = self._sql.execute_fetchone(sql, [username])
-        if row:
-            results = dict(row)
+        with TRN:
+            sql = """SELECT  cast(ag_login_id as varchar(100)) as ag_login_id,
+                      email, name, address, city,
+                      state, zip, country,kit_password
+                    FROM ag_login
+                INNER JOIN ag_kit USING (ag_login_id)
+                WHERE supplied_kit_id = %s"""
+            TRN.add(sql, [username])
+            row = TRN.execute_fetchindex()
+            if row:
+                results = dict(row[0])
 
-            if not bcrypt.verify(password, results['kit_password']):
+                if not bcrypt.verify(password, results['kit_password']):
+                    return False
+                results['ag_login_id'] = str(results['ag_login_id'])
+
+                return results
+            else:
                 return False
-            results['ag_login_id'] = str(results['ag_login_id'])
-
-            return results
-        else:
-            return False
 
     def check_login_exists(self, email):
         """Checks if email for login already exists on system
@@ -142,14 +107,14 @@ class AGDataAccess(object):
         ag_login_id or None
             If exists, returns ag_login_id, else returns None
         """
-        clean_email = email.strip().lower()
-        sql = "select ag_login_id from ag_login WHERE LOWER(email) = %s"
-        cur = self.get_cursor()
-        cur.execute(sql, [clean_email])
-        value = cur.fetchone()
-        if value is not None:
-            value = value[0]
-        return value
+        with TRN:
+            clean_email = email.strip().lower()
+            sql = "SELECT ag_login_id FROM ag_login WHERE LOWER(email) = %s"
+            TRN.add(sql, [clean_email])
+            value = TRN.execute_fetchindex()
+            if not value:
+                return None
+            return value[0][0]
 
     def addAGLogin(self, email, name, address, city, state, zip_, country):
         """Adds a new login or returns the login_id if email already exists
@@ -176,19 +141,19 @@ class AGDataAccess(object):
         ag_login_id : str
             UUID for new user, or existing user if email already in system
         """
-        clean_email = email.strip().lower()
-        ag_login_id = self.check_login_exists(email)
-        if not ag_login_id:
-            # create the login
-            sql = ("INSERT INTO ag_login (email, name, address, city, state, "
-                   "zip, country) VALUES (%s, %s, %s, %s, %s, %s, %s) "
-                   "RETURNING ag_login_id")
-            cur = self.get_cursor()
-            cur.execute(sql, [clean_email, name, address, city,
-                              state, zip_, country])
-            ag_login_id = cur.fetchone()[0]
-            self.connection.commit()
-        return ag_login_id
+        with TRN:
+            clean_email = email.strip().lower()
+            ag_login_id = self.check_login_exists(email)
+            if not ag_login_id:
+                # create the login
+                sql = """INSERT INTO ag_login
+                         (email, name, address, city, state, zip, country)
+                         VALUES (%s, %s, %s, %s, %s, %s, %s)
+                         RETURNING ag_login_id"""
+                TRN.add(sql, [clean_email, name, address, city, state, zip_,
+                              country])
+                ag_login_id = TRN.execute_fetchlast()
+            return ag_login_id
 
     def getAGBarcodeDetails(self, barcode):
         sql = """SELECT  email,
@@ -203,12 +168,13 @@ class AGDataAccess(object):
                   INNER JOIN ag_login USING (ag_login_id)
                   INNER JOIN barcode USING (barcode)
                   WHERE barcode = %s"""
-        row = self._sql.execute_fetchone(sql, [barcode])
 
-        row_dict = {}
-        if row:
-            row_dict = dict(row)
-        return row_dict
+        with TRN:
+            TRN.add(sql, [barcode])
+            row = TRN.execute_fetchindex()
+            if not row:
+                raise ValueError('Barcode does not exist in AG: %s' % barcode)
+            return dict(row[0])
 
     def getAGKitDetails(self, supplied_kit_id):
         sql = """SELECT cast(ag_kit_id as varchar(100)),
@@ -216,12 +182,14 @@ class AGDataAccess(object):
                     kit_verification_code, verification_email_sent
                  FROM ag_kit
                  WHERE supplied_kit_id = %s"""
-        row = self._sql.execute_fetchone(sql, [supplied_kit_id])
 
-        kit_details = {}
-        if row:
-            kit_details = dict(row)
-        return kit_details
+        with TRN:
+            TRN.add(sql, [supplied_kit_id])
+            row = TRN.execute_fetchindex()
+            if not row:
+                raise ValueError('Supplied kit id does not exist in AG: %s' %
+                                 supplied_kit_id)
+            return dict(row[0])
 
     def registerHandoutKit(self, ag_login_id, supplied_kit_id):
         """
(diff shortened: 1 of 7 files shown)


squirrelo:
This is ready for review.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -111,23 +74,25 @@ def authenticateWebAppUser(self, username, password):
...
-                FROM ag_login
-            INNER JOIN ag_kit USING (ag_login_id)
-            WHERE supplied_kit_id = %s"""
-        row = self._sql.execute_fetchone(sql, [username])
-        if row:
-            results = dict(row)
-
-            if not bcrypt.verify(password, results['kit_password']):
+        with TRN:
+            sql = """SELECT  cast(ag_login_id as varchar(100)) as ag_login_id,
```
wasade: Are any of these details used? It looks like the `auth_handler` only checks that the result evaluates `True` or `False`, and I can't see where things like email, name, etc are used.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -142,14 +107,14 @@ def check_login_exists(self, email):
...
-        value = cur.fetchone()
-        if value is not None:
-            value = value[0]
-        return value
+        with TRN:
+            clean_email = email.strip().lower()
+            sql = "SELECT ag_login_id FROM ag_login WHERE LOWER(email) = %s"
+            TRN.add(sql, [clean_email])
+            value = TRN.execute_fetchindex()
+            if not value:
```
wasade: Is `value` a bool?

wasade: oh, I see, it's a list. Could you invert the logic? less thought needed... and then the `return None` is not necessary

amgut/lib/data_access/ag_data_access.py
```diff
@@ -203,12 +168,13 @@ def getAGBarcodeDetails(self, barcode):
...
-        row_dict = {}
-        if row:
-            row_dict = dict(row)
-        return row_dict
+        with TRN:
+            TRN.add(sql, [barcode])
+            row = TRN.execute_fetchindex()
+            if not row:
+                raise ValueError('Barcode does not exist in AG: %s' % barcode)
+            return dict(row[0])
```
wasade: The signature of this method has changed. Previously, an empty `dict` was possible whereas now it raises. Is that expected?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -230,172 +198,154 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
...
+                printresults = 'n'
+            # make sure login_id and skid exists
+            sql = """SELECT EXISTS(SELECT *
+                                   FROM ag.ag_login
+                                   WHERE ag_login_id = %s)"""
+            TRN.add(sql, [ag_login_id])
+            try:
+                exists = TRN.execute_fetchlast()
+            except ValueError:
+                raise ValueError("ag_login_id is not a UUID: %s" % ag_login_id)
```
wasade: This can be explicitly tested prior to entering the context

wasade: i.e., above line 202

amgut/lib/data_access/ag_data_access.py
```diff
@@ -230,172 +198,154 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
...
+                            VALUES (k_id, bc, bc || '.jpg');
+                    END LOOP;
+                    DELETE FROM ag_handout_kits WHERE kit_id = %s;
+                END $do$;
+                """.format(ag_login_id, printresults)
+            TRN.add(sql, [supplied_kit_id] * 3)
+            try:
+                TRN.execute()
+            except psycopg2.IntegrityError:
+                logging.exception('Error on skid %s:' % ag_login_id)
```
wasade: Is that all we know or can the error be more explicit?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -230,172 +198,154 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
...
+                              agc.ag_login_id,
+                              agc.date_signed,
+                              agc.assent_obtainer,
+                              agc.age_range,
+                              agl.survey_id
+                           FROM ag_consent agc
+                           JOIN ag_login_surveys agl
+                           USING (ag_login_id, participant_name)
+                           WHERE agl.survey_id = %s""", [survey_id])
             # Can refactor this with _get_col_names_from_cursor()
```
wasade: is this comment still relevant?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -230,172 +198,154 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
...
-            sql = """SELECT survey_id
-                     FROM ag_login_surveys
-                     WHERE ag_login_id = %s AND participant_name = %s"""
-
-            survey_id = conn_handler.execute_fetchone(
-                sql, (ag_login_id, participant_name))
-            if survey_id:
-                # remove the list encapulation
-                survey_id = survey_id[0]
+        with TRN:
```
wasade: There is a lot of work done within this context, when really, the majority of the logic could be done outside of the context (or contexts). I think the code would be more accessible if the scope of the context only spanned direct interaction with the `TRN`

amgut/lib/data_access/ag_data_access.py
```diff
@@ -453,14 +403,15 @@ def get_vioscreen_status(self, survey_id):
...
-        if status is None:
-            raise ValueError("Survey ID %s not in database" % survey_id)
-        return status[0]
+        with TRN:
+            sql = """SELECT vioscreen_status
+                     FROM ag.ag_login_surveys
+                     WHERE survey_id = %s"""
+            TRN.add(sql, [survey_id])
+            status = TRN.execute_fetchindex()
+            if not status:
```
wasade: This if-block does not need to be part of the context

amgut/lib/data_access/ag_data_access.py
```diff
@@ -481,13 +431,10 @@ def getParticipantSamples(self, ag_login_id, participant_name):
...
-        conn_handler = SQLConnectionHandler()
-        rows = conn_handler.execute_fetchall(
-            sql, [ag_login_id, participant_name])
-        barcodes = [dict(row) for row in rows]
-
-        return barcodes
+        with TRN:
+            TRN.add(sql, [ag_login_id, participant_name])
+            rows = TRN.execute_fetchindex()
+            return [dict(row) for row in rows]
```
wasade: This return does not need to be within the scope of the context

amgut/lib/data_access/ag_data_access.py
```diff
@@ -498,10 +445,10 @@ def getEnvironmentalSamples(self, ag_login_id):
...
                     environment_sampled::text <> '')
                     AND ag_login_id = %s"""
-        rows = self._sql.execute_fetchall(sql, [ag_login_id])
-        barcodes = [dict(row) for row in rows]
-
-        return barcodes
+        with TRN:
+            TRN.add(sql, [ag_login_id])
+            rows = TRN.execute_fetchindex()
+            return [dict(row) for row in rows]
```
wasade: same

amgut/lib/data_access/ag_data_access.py
```diff
@@ -509,80 +456,40 @@ def getAvailableBarcodes(self, ag_login_id):
                  INNER JOIN ag_kit USING (ag_kit_id)
                  WHERE coalesce(sample_date::text, '') = ''
                  AND kit_verified = 'y' AND ag_login_id = %s"""
-        results = self._sql.execute_fetchall(sql, [ag_login_id])
-        return [row[0] for row in results]
+        with TRN:
+            TRN.add(sql, [ag_login_id])
+            return TRN.execute_fetchflatten()
```
wasade: same... will stop commenting on this pattern

amgut/lib/data_access/ag_data_access.py
```diff
@@ -509,80 +456,40 @@ def getAvailableBarcodes(self, ag_login_id):
...
+        with TRN:
+            sql = "SELECT password FROM ag.ag_handout_kits WHERE kit_id = %s"
+            TRN.add(sql, [username])
+            to_check = TRN.execute_fetchindex()
 
-        if not to_check:
-            return False
-        else:
-            return bcrypt.verify(password, to_check[0])
+            if not to_check:
```
wasade: does not need to be part of the context

amgut/lib/data_access/ag_data_access.py
```diff
@@ -694,73 +606,69 @@ def checkPrintResults(self, kit_id):
...
-        else:
-            raise RuntimeError("No user ID for kit %s" % supplied_kit_id)
+        with TRN:
+            sql = """SELECT ag_login_id
+                     FROM ag.ag_kit
+                     JOIN ag_login USING (ag_login_id)
+                     WHERE supplied_kit_id = %s"""
+            TRN.add(sql, [supplied_kit_id])
+            results = TRN.execute_fetchindex()
+            if results:
```
wasade: same

amgut/lib/data_access/ag_data_access.py
```diff
@@ -694,73 +606,69 @@ def checkPrintResults(self, kit_id):
...
-        info = self.getAGKitDetails(supplied_kit_id)
+        with TRN:
+            ag_login_id = self.get_user_for_kit(supplied_kit_id)
+            info = self.getAGKitDetails(supplied_kit_id)
 
-        kit_verified = False
-        if info['kit_verified'] == 'y':
-            kit_verified = True
+            kit_verified = False
+            if info['kit_verified'] == 'y':
```
wasade: same

amgut/lib/data_access/sql_connection.py
```diff
@@ -1,72 +1,204 @@
-from __future__ import division
-
 # -----------------------------------------------------------------------------
-# Copyright (c) 2014--, The American Gut Development Team.
+# Copyright (c) 2014--, The Qiita Development Team.
```
ElDeveloper: Qiita? :L?

amgut/lib/data_access/env_management.py
```diff
@@ -119,25 +119,23 @@ def initialize(verbose=False):
 
 
 def make_settings_table():
-    conn = SQLConnectionHandler()
-    settings = AMGUT_CONFIG.get_settings()
+    with TRN:
+        settings = AMGUT_CONFIG.get_settings()
```
wasade: none of the `settings` interaction needs to be within the context

amgut/lib/data_access/sql_connection.py
```diff
@@ -1,72 +1,204 @@
...
+                      OperationalError)
 from psycopg2.extras import DictCursor
+from psycopg2.extensions import (
+    ISOLATION_LEVEL_AUTOCOMMIT, ISOLATION_LEVEL_READ_COMMITTED,
+    TRANSACTION_STATUS_IDLE)
 
 from amgut.lib.config_manager import AMGUT_CONFIG
 
 
 class SQLConnectionHandler(object):
```
wasade: Can this class be deleted?

wasade:
@josenavas, can you please review as this uses the `Transaction` object?

@squirrelo, general comment is that there is too much work being done within the scope of a context. Like an `if`, `try`, `for` or a function, it is preferential to limit the logic performed in the scope to what is necessary. The benefit is that it improves clarity on the intent of the scope.

amgut/lib/data_access/ag_data_access.py
squirrelo: Yes, more explicit now that something wrong happened, and it can be logged.

amgut/lib/data_access/ag_data_access.py
squirrelo: nope, removed

amgut/lib/data_access/ag_data_access.py
```diff
@@ -176,19 +141,19 @@ def addAGLogin(self, email, name, address, city, state, zip_, country):
...
-                              state, zip_, country])
-            ag_login_id = cur.fetchone()[0]
-            self.connection.commit()
-        return ag_login_id
+        with TRN:
+            clean_email = email.strip().lower()
+            ag_login_id = self.check_login_exists(email)
+            if not ag_login_id:
+                # create the login
+                sql = """INSERT INTO ag_login
```
ElDeveloper: Why not use the addAGLogin method instead of this SQL call?

amgut/lib/data_access/sql_connection.py
josenavas: This code is copy pasted from qiita - I think it is ok leaving this copyright note.

amgut/lib/data_access/ag_data_access.py
squirrelo: My understanding of the idea of the Transaction object is that everything is done within its scope, that way of anything errors, there is a full rollback. Since functions can call other functions wihtin a transaction, that themselves have a transaction, it's safer to just put everything in the context.

amgut/lib/data_access/ag_data_access.py
ElDeveloper: Worthit to add a numpy doc to the method?

amgut/lib/data_access/sql_connection.py
squirrelo: Not yet, as the Survey object relies on it and I don't want to do edits on that without tests in place.

amgut/lib/data_access/ag_data_access.py
josenavas: That's correct. If you want to execute the entire function as a single transaction, this is the correct way of doing it.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -230,172 +198,154 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
...
-
-        try:
-            conn_handler.execute(sql, [supplied_kit_id] * 3)
-        except psycopg2.IntegrityError:
-            logging.exception('Error on skid %s:' % ag_login_id)
-            return False
-        return True
+        with TRN:
+            printresults = self.checkPrintResults(supplied_kit_id)
+            if printresults is False:
```
ElDeveloper: You don't need to check for this n and False will be equivalent in SQL, as long as they are both enclosed by quotes and being inserted into a boolean column - which is this case.

amgut/lib/data_access/sql_connection.py
ElDeveloper: I would be OK with that if we could somehow guarantee that we will not
be modifying this - it's subtle, but it could lead to annoyances in the
future. Maybe worth it to include in the License file that we are
including this file as part of the software? We do this in emperor,
skbio and AFAIK others.

amgut/lib/data_access/ag_data_access.py
wasade: Ah, I see it now. It looked like the if and else (on line 336) where independent blocks of code.

wasade: ...though, the setting of the `sql` string variables is not necessary within the scope

amgut/lib/data_access/ag_data_access.py
```diff
@@ -829,15 +735,15 @@ def get_survey_id(self, ag_login_id, participant_name):
...
-
-        return id_[0] if id_ else None
+        with TRN:
+            sql = """SELECT survey_id
+                     FROM ag_login_surveys
+                     WHERE ag_login_id=%s AND participant_name=%s"""
+            TRN.add(sql, [ag_login_id, participant_name])
+            survey_id = TRN.execute_fetchindex()
+            if not survey_id:
+                raise ValueError("No survey ID found!")
```
ElDeveloper: Can you add a raises section to the docstring?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -803,16 +708,17 @@ def get_login_info(self, ag_login_id):
...
-        return results
+        with TRN:
+            sql = """SELECT ag_login_id, email, name, address, city, state,
+                            zip, country
+                     FROM ag_login
+                     WHERE ag_login_id = %s"""
+            TRN.add(sql, [ag_login_id])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('ag_login_id not in database: %s' %
```
ElDeveloper: Can you add a Raises section to the docstring?

amgut/lib/data_access/ag_data_access.py
squirrelo: They are returned as part of the function (awful yes, but can't change without tests)

amgut/lib/data_access/ag_data_access.py
squirrelo: using `logging.exception` catches the exception and adds it after the message given

amgut/lib/data_access/ag_data_access.py
squirrelo: Done

amgut/lib/data_access/sql_connection.py
squirrelo: I don't think this will get modified, just updated from qiita if necessary. I'll add info to contributing.md about it if wanted.

amgut/lib/data_access/ag_data_access.py
squirrelo: This is addAGLogin?

amgut/lib/data_access/ag_data_access.py
squirrelo: Done

amgut/lib/data_access/ag_data_access.py
squirrelo: Ah, right, this is now a bool column. Thanks!

amgut/lib/data_access/ag_data_access.py
squirrelo: Done

amgut/lib/data_access/ag_data_access.py
squirrelo: Done

squirrelo:
Changes are added, not sure if we are now ignoring all the context questions as it goes against how the Transaction object was planned to be used.

wasade:
Thanks. I guess maybe over recommending re exceptions? Sorry

amgut/lib/data_access/ag_data_access.py
```diff
@@ -216,186 +200,181 @@ def getAGKitDetails(self, supplied_kit_id):
...
+            row = TRN.execute_fetchindex()
+            if not row:
+                raise ValueError('Supplied kit id does not exist in AG: %s' %
+                                 supplied_kit_id)
+            return dict(row[0])
 
     def registerHandoutKit(self, ag_login_id, supplied_kit_id):
-        """
+        """Registeres a handout kit to a user
+
```
mortonjt: Registers?

amgut/lib/data_access/sql_connection.py
mortonjt: :+1:

amgut/lib/data_access/ag_data_access.py
```diff
@@ -216,186 +200,181 @@ def getAGKitDetails(self, supplied_kit_id):
...
+                     WHERE ag_login_id = %s AND participant_name = %s"""
+            TRN.add(sql, (ag_login_id, participant_name))
+            survey_id = TRN.execute_fetchlast()
 
-        with conn_handler.get_postgres_cursor() as curr:
-            sql = ("DELETE FROM survey_answers WHERE "
-                   "survey_id = %s")
-            curr.execute(sql, [survey_id])
+            sql = "DELETE FROM survey_answers WHERE survey_id = %s"
+            TRN.add(sql, [survey_id])
```
mortonjt: I'm seeing a lot of repetitive code.  It may be preferable to define helper methods, to minimize future modifications.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -216,186 +200,181 @@ def getAGKitDetails(self, supplied_kit_id):
...
 
-            sql = ("DELETE FROM survey_answers_other WHERE "
-                   "survey_id = %s")
-            curr.execute(sql, [survey_id])
+            sql = "DELETE FROM survey_answers_other WHERE survey_id = %s"
+            TRN.add(sql, [survey_id])
```
mortonjt: Here's another example of an identical pattern

amgut/lib/data_access/ag_data_access.py
```diff
@@ -216,186 +200,181 @@ def getAGKitDetails(self, supplied_kit_id):
...
             sql = "DELETE FROM promoted_survey_ids WHERE survey_id = %s"
-            curr.execute(sql, [survey_id])
+            TRN.add(sql, [survey_id])
 
             # Delete last due to foreign keys
-            sql = ("DELETE FROM ag_login_surveys WHERE "
-                   "survey_id = %s")
-            curr.execute(sql, [survey_id])
+            sql = "DELETE FROM ag_login_surveys WHERE survey_id = %s"
+            TRN.add(sql, [survey_id])
```
mortonjt: Same here.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -216,186 +200,181 @@ def getAGKitDetails(self, supplied_kit_id):
...
-                   "survey_id = %s")
-            curr.execute(sql, [survey_id])
+            sql = """UPDATE ag_kit_barcodes
+                     SET survey_id = NULL
+                     WHERE survey_id = %s"""
+            TRN.add(sql, [survey_id])
 
             sql = "DELETE FROM promoted_survey_ids WHERE survey_id = %s"
-            curr.execute(sql, [survey_id])
+            TRN.add(sql, [survey_id])
```
mortonjt: Same here.

amgut/lib/data_access/ag_data_access.py
mortonjt: :+1:

mortonjt:
Question - since the `sql_connection.py` is being ported over from qiita, would it be worthwhile to also port over the tests associated with `sql_connection.py`?

mortonjt:
Also, not sure what is going on with the build - it seems to have stalled even after I restarted it.

amgut/lib/data_access/ag_data_access.py
squirrelo: That's how the SQL works though. You write the SQL, then run it. Not sure it can be decomposed any more than that.

amgut/lib/data_access/ag_data_access.py
squirrelo: Yes, but this is subtly different than the above one: it's acting on a different table. You would need to define a TON of helper methods to do what you are suggesting (plus joins etc complicate things further) so I don't think that is the best approach.

wasade: Well, one possibility would be something like:

```python
def one_arg(sql, arg):
    with TRN:
        TRN.add(sql, [arg])
```

...or something, which would reduce some code but it does obfuscate some as well

amgut/lib/data_access/ag_data_access.py
wasade: @josenavas, I can't find the docs for the `Transaction` object in qiita short of what's in code. Could you point @mortonjt at them? @mortonjt, the intent of the object is to automatically rollback if any issue occurs. It is particularly useful when needing to execute multiple statements that cannot be combined into 1 easily (and a procedure doesn't make sense), or in cases where a mix of SQL and Python interaction is necessary, and where either the SQL or the Python may trigger a "failure".

amgut/lib/data_access/sql_connection.py
wasade: That's assuming this code is maintenance free...

wasade:
Tests stalled again. @squirrelo, can you add a verbose flag into `nosetests` on `.travis.yml`? It'll show us what test is blocking

amgut/lib/data_access/ag_data_access.py
mortonjt: You're looking for this [file](https://github.com/biocore/qiita/blob/master/qiita_db/sql_connection.py) right?

amgut/lib/data_access/ag_data_access.py
mortonjt: Yeah, you really only need one simple helper function, namely
```
def delete_stuff(table, survey_id):
    with TRN:
        sql = "DELETE FROM %s WHERE survey_id = %s"
        TRN.add(sql, [table, survey_id])
```

squirrelo: only a few deletes are `survey_id`. To do this generally, we would need what @wasade showed, which in my opinion sacrifices a lot of readability for not a lot of code savings. What do we gain by breaking this out as opposed to leaving it as is? To me this just adds more maintenance and obfuscation, since the SQL language will (presumably) not change how base calls are made.

josenavas: @mortonjt in this case I'm with @squirrelo.
If there is a lot of SQL being obscured in functions, it is really easy to end up modifying something that we are not supposed to. Also, @wasade's proposal will not buy anything else compared with what the Transaction object already provides.

amgut/lib/data_access/sql_connection.py
ElDeveloper: I see, it would make sense to mention this in the licensing file.

amgut/lib/data_access/ag_data_access.py
ElDeveloper: haha, my bad!

squirrelo: :laughing:

squirrelo:
Found the issue, and licence stuff added, should be good now.

mortonjt:
I still have a few comments that aren't unanswered.  Otherwise, I think it looks ok.

squirrelo:
Tests ported, so good to go.

mortonjt:
:+1:

Issue #526, opened 2015-10-02
Add tests for handlers


squirrelo:
- [ ] add_sample.py
- [ ] add_sample_overview.py
- [ ] addendum.py
- [ ] animal_survey.py
- [ ] auth_handlers.py
- [ ] base_handlers.py
- [ ] change_pass_verify.py
- [ ] change_password.py
- [ ] download.py
- [ ] forgot_password.py
- [ ] help_request.py
- [ ] human_survey.py
- [ ] human_survey_completed.py
- [ ] international.py
- [ ] new_participant.py
- [ ] new_participant_overview.py
- [ ] nojs.py
- [ ] open_humans.py
- [ ] participant_overview.py
- [ ] portal.py
- [ ] retrieve_kitid.py
- [ ] sample_overview.py
- [ ] survey.py
- [ ] taxa_summary.py
- [ ] vioscreen.py

PR #529, opened 2015-10-06
Test add_sample.py

squirrelo:
This adds the first tests for tornado to american gut, in the form of the add_sample pages. This also found a lot of bugs, so the ones that were caught were fixed as well.


squirrelo:
Ready for review.

squirrelo:
Part of #526

mortonjt:
:+1:

PR #535, opened 2015-10-08
Test survey

squirrelo:
This adds all tests to the survey object.

### DIFF
--- a/amgut/lib/data_access/test/test_survey.py
+++ b/amgut/lib/data_access/test/test_survey.py
@@ -1,0 +1,96 @@
+from unittest import TestCase, main
+from wtforms.form import BaseForm
+from amgut.lib.data_access.survey import (
+    Question, QuestionSingle, QuestionMultiple, QuestionText, QuestionString,
+    Group, Survey)
+
+
+class TestQuestionSingle(TestCase):
+    def test_create_no_triggers(self):
+        q_single = QuestionSingle(23, 'Education')
+        self.assertEqual(q_single.id, 23)
+        self.assertEqual(q_single.group_name, 'Education')
+        self.assertEqual(q_single.set_response, None)
+        self.assertEqual(q_single.response_type, 'SINGLE')
+        self.assertEqual(q_single.question,
+                         'What is your highest level of education?')
+        self.assertEqual(q_single.american_question,
+                         'What is your highest level of education?')
+        self.assertEqual(q_single.triggers, tuple())
+        self.assertEqual(q_single.qid, 'Education_23')
+        self.assertEqual(q_single.interface_element_ids, ['Education_23_0'])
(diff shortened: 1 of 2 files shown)


amgut/lib/data_access/test/test_survey.py
```diff
@@ -0,0 +1,452 @@
...
+               'Pet_Information_145_0': '29 - Male',
+               'Pet_Information_136_0': 3,
+               'Pet_Information_135_0': [3],
+               'Pet_Information_141_0': 4}
+        self.assertEqual(obs, exp)
+
+        obs = ag_data.getConsent(survey_id)
+        consent['ag_login_id'] = consent['login_id']
+        del consent['login_id']
+        del consent['obtainer_name']
```
mortonjt: The `del` statements are a little bothersome.  Personally, I'd prefer to create a completely new object, since its already be heavily modified.

But this is just personal preference - definitely not blocking

amgut/lib/data_access/test/test_survey.py
```diff
@@ -0,0 +1,452 @@
...
+        without_fk = [(survey_id, 130, '["20"]'),
+                      (survey_id, 142, '["Giant ratty pet!"]'),
+                      (survey_id, 143, '["Capybara"]'),
+                      (survey_id, 144, '[""]'),
+                      (survey_id, 145, '["29 - Male"]'),
+                      (survey_id, 127, '["%s"]' % name)]
+
+        survey.store_survey(consent, with_fk, without_fk)
+
+        obs = survey.fetch_survey(survey_id)
```
mortonjt: Not within the scope of this PR.

But would it be desirable to enforce return types from `fetch_survey`?  I see that there are lists, strings and ints that all can be observed.

amgut/lib/data_access/test/test_survey.py
squirrelo: The dels are so the same dictionary that was used to add the info is used to test what comes out, minus the columns not pulled out. This way I can assure in == out.

mortonjt:
First pass through, looking good.

Now would this be an appropriate time to incorporate tests for UTF?  Or should that be tested later?

amgut/lib/data_access/test/test_survey.py
squirrelo: The return types are from the database, so I'm not sure what you mean? The lists are from multiple answer questions or free text, and the rest are from controlled vocab that is already typed in the database.

squirrelo:
I can add umlauts to the store survey to make sure UTF-8 works. Good catch.

squirrelo:
umlauts added. Test is now 100% more metal :metal:

mortonjt:
:+1:

PR #536, opened 2015-10-08
Remove SQLConnectionHandler

squirrelo:
This updates the survey to use the Transaction object. Since that was the last thing using the SQLConnectionHandler for database access, it also removes that as well.

Requires #535


amgut/lib/data_access/survey.py
```diff
@@ -34,51 +32,56 @@ class Question(object):
...
+                         ON qr.response = sr.{0}
+                     WHERE q.survey_question_id = %s
+                     ORDER BY qr.display_index
+                  """.format(_LOCALE_COLUMN, self._survey_question_table,
+                             self._question_response_table,
+                             self._response_table)
+            TRN.add(sql, [self.id])
+            self.responses = TRN.execute_fetchflatten()
+
+            if not self.responses:
```
mortonjt: What sort of things would trigger this call?

squirrelo: String or Text question type which doesn't have set responses

mortonjt:
Don't forget to resolve the merge conflicts

squirrelo:
Resolved, good to review

amgut/lib/data_access/survey.py
```diff
@@ -282,50 +294,46 @@ def fetch_survey(self, survey_id):
...
+            answers_other = TRN.execute_fetchindex()
+
+            survey = defaultdict(list)
+            for qid, idx, qtype in answers:
+                eid = self.questions[qid].interface_element_ids[0]
+                if qtype == 'SINGLE':
+                    survey[eid] = idx
+                else:
+                    survey[eid].append(idx)
+
```
mortonjt: This is another personal preference

I'm not a big fan of hardcoding numbers.  Is it possible to replace these numbers with constants, or add a comment indicating why those indices are chosen.

squirrelo: Byproduct of the Transaction object, so can't be changed.

mortonjt: Sorry, was referring to line 330

amgut/lib/data_access/survey.py
```diff
@@ -282,52 +294,46 @@ def fetch_survey(self, survey_id):
...
+            for qid, data in answers_other:
+                eid = self.questions[qid].interface_element_ids[0]
+                data = str(data[2:-2])  # drop [""]
```
mortonjt: Here's what I was referring to

mortonjt:
One small comment, but otherwise looks good

amgut/lib/data_access/survey.py
squirrelo: There is a comment at the end of the line explaining what it does? I'm not sure what else I can do.

squirrelo: Also note I did not program that, so it's been there since the beginning.

squirrelo: although looking at that, I think there is a cleaner way to do it. Will futz with it in a bit.

squirrelo:
Comment addressed, good to go.

mortonjt:
:+1:

PR #539, opened 2015-10-15
Open Humans connection process update

beaugunderson:
This pull request contains updates for the Open Humans connections process, mostly aimed at providing a better user experience for that process.

It's nearly ready to merge (I'd like to do a bit more local testing) but it's ready for feedback. :sparkles:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,3 +1,6 @@
 [flake8]
 exclude = amgut/lib/locale_data/*,amgut/templates/*,amgut/static/*,amgut/db/*
 max-complexity = 10
+
+[pep257]
+ignore = D100,D102,D200,D203,D204,D205,D400
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.eggs/
 
 # Installer logs
 pip-log.txt
--- a/amgut/handlers/open_humans.py
+++ b/amgut/handlers/open_humans.py
@@ -1,6 +1,4 @@
 import logging
-import posixpath
-import urlparse
 
 try:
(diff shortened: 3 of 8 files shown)


wasade:
I don't see any issues. @squirrelo, can you review too?

tox.ini
```diff
@@ -1,3 +1,6 @@
 [flake8]
 exclude = amgut/lib/locale_data/*,amgut/templates/*,amgut/static/*,amgut/db/*
-max-complexity = 10
\ No newline at end of file
+max-complexity = 10
+
+[pep257]
+ignore = D100,D102,D200,D203,D204,D205,D400
```
squirrelo: Not sure if we should be ignoring all these, and if we add this section are we going to then start enforcing pep257 on this repo?

squirrelo:
Code looks good to me. Just have 1 question about pep257 setup.

tox.ini
beaugunderson: I had my own `.pep257` file in the directory for the code I was adding; I set it to ignore some of the more pedantic errors. `pep257` is unfortunately written in a way that if there's a `tox.ini` file in the directory that it always looks for its settings there and no longer looks in the `.pep257` file.

Decision about whether you all want to use it or not is up to you. :)

wasade:
@squirrelo, if you're :+1:, can you merge?

squirrelo:
Still says it's WIP

beaugunderson:
@squirrelo just removed the WIP, I've tested it on our end as much as I can :)

squirrelo:
Code looks good, so merging.

beaugunderson:
@squirrelo for planning purposes on our end do you have an idea when this will be deployed to production? :)

squirrelo:
Should be now :boom:

PR #542, opened 2015-10-28
Add patch for duplicates tracking

squirrelo:
This adds a table for tracking duplicate survey participant names and populates it.


amgut/db/patches/0029.sql
```diff
@@ -0,0 +1,174 @@
+-- Oct 27, 2015
+-- Add way to track what surveys are duplicates for participant consent
+
+--Add timestamp for when a kit is registered
+ALTER TABLE ag.ag_kit ADD registered_on timestamp DEFAULT current_timestamp;
+
+--Add table to track duplicates
+CREATE TABLE ag.duplicate_consents (
```
wasade: no explicit pk? the indexing scheme below suggests either having a `unique` constraint over all three columns, or to create an aggregate natural primary key of all three columns. what do you think?

squirrelo: agreagate PK would work best, with a unique on duplicate survey. Will do that now.

wasade: thanks!

squirrelo:
Change made, ready to go/

PR #543, opened 2015-10-31
Add explicit delete of registered handout barcodes

squirrelo:
Still getting an issue of barcodes not being deleted when registered using the cascade. This makes it explicit.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -276,7 +276,7 @@ class AGDataAccess(object):
                     DELETE FROM ag_handout_kits WHERE kit_id = %s;
                 END $do$;
                 """.format(ag_login_id, printresults)
-            TRN.add(sql, [supplied_kit_id] * 3)
+            TRN.add(sql, [supplied_kit_id] * 4)
             try:
                 TRN.execute()
             except psycopg2.IntegrityError:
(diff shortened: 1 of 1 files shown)


squirrelo:
Anyone want to review this? It's a really small PR.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -272,10 +272,11 @@ def registerHandoutKit(self, ag_login_id, supplied_kit_id):
                             (ag_kit_id, barcode, sample_barcode_file)
                             VALUES (k_id, bc, bc || '.jpg');
                     END LOOP;
+                    DELETE FROM ag_handout_barcodes WHERE kit_id = %s;
                     DELETE FROM ag_handout_kits WHERE kit_id = %s;
                 END $do$;
                 """.format(ag_login_id, printresults)
-            TRN.add(sql, [supplied_kit_id] * 3)
+            TRN.add(sql, [supplied_kit_id] * 4)
```
mortonjt: Can you add a comment about where the number 4 comes from?

squirrelo: The %s from above?

mortonjt: It changed from 3 to 4.  Any reason why?
And could you add that reason in a comment.  Otherwise, I think it is ready to merge.

mortonjt: Oh, I see nvm.

PR #547, opened 2015-11-11
WIP - Plate mapper DB

squirrelo:
This is the inital design for the plate mapper and how data will be stored. Please take a close, close look at this and ask as many questions as needed. The design should be extensible enough to cover not just what we currently do in the lab, but anything from metabolomics to WGS and onwards. 

Due to the complexity and importance of this PR, I'd like to get at least 3 people to review it, if not more. This is currently for design decisions. Once we agree on the design, I will generate the SQL using DBSchema and add it to this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ ag_config.txt
 *.swp
 *.swo
 *~
+*.bak
 
 # Foreman
 .env
--- a/amgut/db/ag.html
+++ b/amgut/db/ag.html
@@ -96,7 +96,7 @@ table {
 </head>
 
 <body>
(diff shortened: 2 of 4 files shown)


squirrelo:
ping @biocore/american-gut-admin @biocore/american-gut-web-admin

wasade:
Why is this in AG web and not labadmin?

squirrelo:
DB changes need to happen in AG since we can't patch from labadmin. All subsequent things will be in labadmin.

squirrelo:
No comments, so closing to make final PR

PR #557, opened 2015-11-30
Plate mapper db changes

squirrelo:
This adds in the tables and database modifications needed for a full sample tracking and prep information storage and retrieval.

This needs to be in AG because all DB patching is done through AG. All subsequent code PRs will be in labadmin.


squirrelo:
Ready for review. Incorporates the few comments I got on the original pull request.

mortonjt:
First pass through.  I don't see any obvious problems.

However, given that there is a ton of changes to the database, I would recommend against merging this without any additional unit-tests.

squirrelo:
There is no way to unit test any of this, as it has no data and the structures using it still have not been added to labamin.

squirrelo:
Had some discussions and made a few more changes, this should now be the final version and ready to merge.

mortonjt:
Sounds good

PR #565, opened 2016-01-12
More log samples messages

squirrelo:
This adds a bold, red message to the top of each portal tab stating there are samples that were received but do not have consent. It also adds a reminder at the end of the survey that you need to associate a sample in order to get it processed.


squirrelo:
Here is the message in context: 
screen shot 2016-01-12 at 16 38 33

EmbrietteH:
Really minor point: can not = cannot
Also, would it be appropriate to perhaps add a short phrase like: "A list of non-associated barcodes appears below"?
ping @mortonjt @JWDebelius for a second set of eyes

mortonjt:
The rendering looks good. I think it is good to go once @EmbrietteH 's comment is addressed

squirrelo:
Change made, ready to go.

PR #566, opened 2016-01-30
Small fix for Open Humans

beaugunderson:
Not sure what changed (I looked at recent commits and didn't see anything that would impact this) but @madprime noticed that people linked American Gut to Open Humans were having their participant ID saved and not their survey ID. This did work at one point, though maybe my code relied on something from the database that it shouldn't have. This commit makes it explicit that it's the `survey_id` we're sending.

cc @EmbrietteH

### DIFF
--- a/amgut/static/js/open_humans.js
+++ b/amgut/static/js/open_humans.js
@@ -9,7 +9,7 @@ $(function () {
 
   $('#connect-open-humans').click(function () {
     var participants = $('.participant:checked').map(function () {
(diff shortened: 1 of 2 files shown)


beaugunderson:
cc'ing @squirrelo for deployment help :)

squirrelo:
:+1: for our end. Are there any other updates that need to be done during deploy?

beaugunderson:
nope, nothing else needed :)

---

PR #573, opened 2016-02-29
Also null out survey id when removing barcode

squirrelo:
This fixes an oversight when removing the barcode where the survey ID was left in despite the barcode being un-associated with anything.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -386,7 +386,8 @@ class AGDataAccess(object):
             sql = """UPDATE ag_kit_barcodes
                      SET participant_name = NULL, site_sampled = NULL,
                          sample_time = NULL, sample_date = NULL,
-                         environment_sampled = NULL, notes = ''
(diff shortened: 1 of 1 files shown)


amgut/lib/data_access/ag_data_access.py
```diff
@@ -386,7 +386,8 @@ def deleteSample(self, barcode, ag_login_id):
             sql = """UPDATE ag_kit_barcodes
                      SET participant_name = NULL, site_sampled = NULL,
                          sample_time = NULL, sample_date = NULL,
-                         environment_sampled = NULL, notes = ''
+                         environment_sampled = NULL, notes = NULL,
```
wasade: is the change in type associated with `notes` going to be problematic?

squirrelo: No, notes can be null, so this is fixing it so that it is consistent with a non-logged barcode instead of slightly different with the empty string.

wasade: :+1:

PR #570, opened 2016-02-11
Db changes

squirrelo:
This makes the changes necessary to track redcap surveys in the AG database.

PR #579, opened 2016-03-17
Redcap survey links

squirrelo:
Adds helper functions and config options for back and forth communication with redcap API
~~Requires #570~~ Merged


amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,115 @@
...
+from amgut.lib.config_manager import AMGUT_CONFIG
+from amgut.lib.data_access.sql_connection import TRN
+
+
+def _make_request(data):
+    body = urlencode(data)
+    client = HTTPClient()
+    response = client.fetch(AMGUT_CONFIG.redcap_url, method='POST',
+                            headers=None, body=body)
+    # Response not always JSON, so can't always use loads
```
wasade: This comment does not reflect the code as a `loads` is used on line 18. Or is a valid response one that is not JSON?

amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,115 @@
...
+        Redcap record (consent) to attach this survey to
+    instrument : str, optional
+        The instrument (survey) to get the URL for. Default ag_human_en_us
+
+    Returns
+    -------
+    str
+        URL for the survey
+    """
+    with TRN:
```
wasade: What's to stop duplicate redcap_event_ids from being returned, or is that acceptable? The concern is it looks like this pattern is for an autoincrementing key, but the transaction is not paired with an insert to "block" the key returned.

amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,115 @@
...
+    Parameters
+    ----------
+    records : list of int
+        record to log for
+    instrument : str
+        The instrument (survey) to log for
+    event : int
+        Event to log for
+    """
+    with TRN:
```
wasade: Ya, so, the insert is in `log_complete`. It is possible that multiple executions of `get_survey_url` can result in the same `redcap_event_id`. If/when that happens, this code will except assuming that `redcap_instrument_id` and `redcap_record_id` form a primary key

amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,115 @@
...
+    Parameters
+    ----------
+    records : list of int
+        records to pull information for
+    instrument : str
+        The instrument (survey) to get responses from
+    event : int, optional
+        What event to get responses from. Default all
+
+    Returns: list of dict of objects
```
wasade: numpydoc structure

amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,115 @@
...
+        'rawOrLabel': 'label',
+        'rawOrLabelHeaders': 'label',
+        'exportCheckboxLabel': 'true',
+        'exportSurveyFields': 'false',
+        'exportDataAccessGroups': 'false',
+        'returnFormat': 'json'
+    }
+    if event is not None:
+        data['events'] = 'event_%d_arm_1' % event,
+    return loads(_make_request(data))
```
wasade: serialize and the deserialize?

wasade:
Few comments

amgut/lib/data_access/redcap.py
squirrelo: Only valid responses can be non-JSON. if there is an error, it will always be JSON and have the error key. The other responses can be either JSON or non-JSON. For example, the survey url comes in as plain text, but the responses information comes in as JSON.

amgut/lib/data_access/redcap.py
squirrelo: _make_request returns the raw string of response.body, and this call to redcap returns JSON. Because of this, I need to deserialize the raw string to return the dict.

amgut/lib/data_access/redcap.py
squirrelo: The code in the next PR actually expects that the same redcap_event_id is used if a survey is not completed. When someone starts a survey but doesn't complete it, it is not logged in redcap nor the AG db. This function is called as part of the AG passthrough page, which is where participants are sent by redcap after their survey is logged. This allows us to not waste event-ids on unfinished surveys and still track what surveys people have taken.

amgut/lib/data_access/redcap.py
squirrelo: fixed

amgut/lib/data_access/redcap.py
wasade: Aren't there like 2^64 possible event IDs? I don't know if we should be concerned about waste given the size of the namespace. 

Are you indicating that duplicate `(redcap_instrument_id, redcap_event_id)` tuples are not a problem?

amgut/lib/data_access/redcap.py
wasade: Just to clarify:

* a valid response is not JSON
* an invalid response is JSON

What about:

```
try:
    error = loads(response.body)['error']
except:
    pass
else:
    raise HTTPError(400, error)
```

amgut/lib/data_access/redcap.py
wasade: Oh, I see now, thanks

amgut/lib/data_access/redcap.py
squirrelo: That block is not in place because people can start the survey but not complete it, leaving an event id in limbo. The only scenario I can think of where someone would have a duplicate event_id is if someone has two browsers open and logged into the same person. This call is made internally in place of the original survey start, and automatically redirects to the survey URL. If blocking the event ID is a concern, I can try and come up with something that clears blocked event_ids after a couple hours or so,  but this may cause a leapfrog effect where someone starts a survey, then doesn't finish it, then starts a survey again before the clear and loses the blocked event_id.

amgut/lib/data_access/redcap.py
squirrelo: Each event ID needs to be manually set up in redcap, so while I could do 2^64, that might take a bit :) The manual nature of redcap is why I'm concerned about the duplictes, but if someone sits for an hour or two and adds in a few thousand events I think the blocker could work. Retrieval code would need to be changed, however.

squirrelo: Also, to clarify, that's per survey as well, so animal, human, and eventually environment would all need those manual events added separately, as well as each survey in a differing language.

amgut/lib/data_access/redcap.py
squirrelo: A valid response may be JSON, or it may not be, depending on what is being queried. That's why this is complicated. However, it will only have the error key if it errors, so the try except suggested should still work.

amgut/lib/data_access/redcap.py
wasade: I don't see that from the query. It looks to me as though two separate people, if they request a new survey at approximately the same time, will receive the same `redcap_event_id` since `redcap_event_id` is only incremented when a survey is completed.

@josenavas @antgonza, if you have a moment, would it be possible to comment? 

@squirrelo, it's possible there is something fundamental about redcap that I do not understand...

antgonza: @wasade I think you are right, same id to multiple clients + this doesn't update the redcap_event_id, right? Perhaps the solution is to have a redcap_event_id per user, not sure?

squirrelo: Ah, ok, I think I see the confusion now. the redcap record ID is what ties a single person to the surveys. Think of it as the host_subject_id, and it is attached to the consent. the event_id is the actual survey ID, and is attached to a single instance of someone taking the survey. Think of it as a row of metadata.

By passing the record into this function, you assure that a single user will not overlap event IDs. If two separate people request at the same time, they have differing record IDs, so the same event ID doesn't matter.

amgut/lib/data_access/redcap.py
wasade: Wait, so you _are_ concerned about duplicates? 

What I'd like to do is resolve a) if it is possible for duplicates IDs to be generated and b) if it is a problem if duplicate IDs are generated. If the answer to both of those is yes, then I do not think the current strategy is safe.

2^64 is approximately 10^22. So, even if every person on the planet filled out a survey, and each survey generated 1000 IDs, we'd still have comfortable breathing room.

amgut/lib/data_access/redcap.py
squirrelo: Also, event_ids count up starting from 1 for each record in redcap, so this code takes advantage of that.

amgut/lib/data_access/redcap.py
wasade: okay, thanks

amgut/lib/data_access/redcap.py
wasade: Ah, okay. Would it be possible to add documentation either to the method, or perhaps at the top of the file which describes how to interpret the IDs? I worry that this could easily be confusing again 6 months from now

josenavas: I'm unsure if I'm understanding this correctly or not. I think what is happening here is that this code is trying to predict the survey ID that redcap will create for the given user. Wouldn't it make more sense to actually ask to redcap which is the new ID? That way, if redcap one day changes and is no longer auto-incremented, our code will not need to change.

squirrelo: I can do an explanation at the top of how redcap breaks down their IDs and other pertinent information. I'll add that now.

wasade: Thanks!

squirrelo: There is no prediction going on. The survey ID is set at survey creation, and the events are all manually created per survey so we know what the event IDs will be.

squirrelo:
Comments addressed and explanation of redcap added to the top.

antgonza:
Thanks, I think there is only one outstanding question from @josenavas. Why are we generating the ids vs. requesting redcap to generate them in get_survey_url?

wasade:
@antgonza, the IDs are generated prior to the system going live. They are static artifacts of the system. I also found this confusing until chatting with @squirrelo, but he is adding in a detailed docstring about it.

amgut/lib/data_access/redcap.py
```diff
@@ -0,0 +1,159 @@
...
+# Redcap has three layers of IDs: record_id, survey_id, and event_id.
+
+# record_id is the id given to a single person in the redcap system, analagous
+# to the host_subject_id.
+
+# survey_id is the name of the survey the person will be taking.
+
+# event_id is the id of the single time a person has taken the survey. It
+# counts up from 1 for each person in each survey. These must be manually
+# created for each survey before the survey is used, so we don't want to waste
```
wasade: Is the following accurate?

"An event_id is the ID of a single time a person has taken a survey. It corresponds to a single instance of a survey for a given survey type and user. Within REDCap, it is necessary to create these "instance" IDs in advance, and are static artifacts of the system. These IDs can be used to distinguish separate surveys that an individual has taken (e.g., multiple time points)."

squirrelo: better wording for the last sentence would be `distinguish separate rounds of the same survey that an individual has taken...` but other then that yes.

wasade: Ah, yes.

@josenavas @antgonza, does this make sense?

@squirrelo, sorry for focusing so much on this. It is obviously generating confusion so just want to make sure we work through it as cleanly as we can so we hate ourselves less in the future

squirrelo: No prob, It's a confusing system at first with a lot of layers, so good to have it explained first go around then later.

antgonza:
Thanks guys!

PR #582, opened 2016-04-08
Add emperor files

squirrelo:
This adds all the support files for emperor into the repo. There is no new code in this pull request.


squirrelo:
@EmbrietteH @ElDeveloper @mortonjt This is just a copy-paste of the emperor support files so can be merged fast.

PR #591, opened 2016-05-31
New surveys

squirrelo:
This adds the fermented foods and surfing surveys into the possible secondary surveys for humans. 

This also adds the ability to log multiple surveys to the same existing consent.

Requires #590

PR #593, opened 2016-06-08
Auto-withdraw or release barcodes

squirrelo:
This updates the code for someone revoking consent to have the barcode either returned to be re-associated if the person still has it, or set to withdrawn if the sample has already been sent in. It also adds logic to log when a participant has withdrawn consent.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -296,11 +296,12 @@ class AGDataAccess(object):
     def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
         # Remove user from new schema
         with TRN:
-            sql = """SELECT survey_id
+            sql = """SELECT survey_id, email
                      FROM ag_login_surveys
+                     JOIN ag_consent USING (ag_login_id, participant_name)
                      WHERE ag_login_id = %s AND participant_name = %s"""
             TRN.add(sql, (ag_login_id, participant_name))
-            survey_id = TRN.execute_fetchlast()
+            survey_id, participant_email = TRN.execute_fetchlast()
 
             sql = "DELETE FROM survey_answers WHERE survey_id = %s"
             TRN.add(sql, [survey_id])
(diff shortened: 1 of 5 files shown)


EmbrietteH:
ping @josenavas @wasade

amgut/lib/data_access/ag_data_access.py
```diff
@@ -380,25 +400,52 @@ def logParticipantSample(self, ag_login_id, barcode, sample_site,
...
+            Barcode to delete
+        ag_login_id : UUID4
+            Login ID for the barcode
+
+        Notes
+        -----
         Strictly speaking the ag_login_id isn't needed but it makes it really
         hard to hack the function when you would need to know someone else's
-        login id (a GUID) to delete something maliciously
+        login id (a GUID) to delete something maliciously,
```
josenavas: `,` -> `.`

amgut/lib/data_access/sql_connection.py
```diff
@@ -53,6 +53,7 @@ def __init__(self):
         self._connection = None
         self._post_commit_funcs = []
         self._post_rollback_funcs = []
+        self.auto_rollback = False
```
josenavas: I disagree with this change. It looks it is added just because is needed in the test that you're adding below. Instead of modifying this library, that should be the same used in Qiita, you can use the `tearDown` function in the tests. Having small modifications like this in code that is similar (or the same) across projects will make code maintenance harder.

josenavas:
Some comments

amgut/lib/data_access/sql_connection.py
squirrelo: The problem is `deleteAGParticipantSurvey` makes a lot of changes to the database, and since we are committed to only using access functions and not direct SQL in the tests, we can't set up a tearDown that won't error on every other test. It makes more sense to just rollback after the test, which also ensures 100% that the DB goes back to before the tests.

josenavas: what about a reset of the test database? It will be consistent with the other projects.

squirrelo: That would add 45 seconds to a minute per test, and was agreed upon with @wasade when we first started adding tests would not be done.

josenavas: Ok, how is that handled in other parts of the test? There are no tests that are modifying the database?

As an alternative, you can add a new element that you are going to delete in the test.

squirrelo: All the tests are modifying the database, but were designed in such a way as to either use other functions to reset as part of the testing of both functions or other ways of making sure the tests can be run multiple times.

squirrelo:
Tests pass, but need #591 merged to get the test_util.py file to add the rollback test to so there are no merge conflicts later. That makes this dependent on #591 now.

josenavas:
@squirrelo can you pull from upstream and fix tests in here? Thanks!

squirrelo:
Test added

josenavas:
👍

PR #599, opened 2016-06-21
Centralize participant_name info

squirrelo:
This removes the legacy participant_name column in ag_kit_barcodes and fixes all SQL that was using it.


squirrelo:
This is a major bug and needs to be merged before deployment.

josenavas:
👍

PR #604, opened 2016-06-30
Update open humans survey for get_survey_id changes

squirrelo:
Requires #603 
Updates the open humans interface code to send information on the main human survey to open humans.
@beaugunderson please review if you can, since this is open humans.

### DIFF
--- a/amgut/handlers/open_humans.py
+++ b/amgut/handlers/open_humans.py
@@ -58,8 +58,8 @@ class OpenHumansHandler(BaseHandler, OpenHumansMixin, OriginMixin):
             survey_ids = {}
 
             for participant_name in human_participants:
-                survey_id = ag_data.get_survey_id(ag_login_id,
-                                                  participant_name)
+                survey_id = ag_data.get_survey_ids(ag_login_id,
+                                                   participant_name)[0]
 
                 if survey_id:
                     survey_ids[participant_name] = survey_id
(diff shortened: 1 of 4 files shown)


beaugunderson:
@squirrelo looks good to me, thanks! :)

EmbrietteH:
Comment from Madeleine: 

Why is this line [1]:
https://github.com/squirrelo/american-gut-web/blob/open-humans-survey_id/amgut/handlers/participant_overview.py#L43

But this one is [0]?
https://github.com/squirrelo/american-gut-web/blob/open-humans-survey_id/amgut/handlers/open_humans.py#L62

I'm asking because I want to check that we will be getting the correct data. Are we going to continue getting a survey ID that can be used to match to the corresponding sample within EBI data?

Currently we're able to match and retrieve FASTQ data from EBI based on survey ID, which is within the metadata for each sample. For example, this sample has survey ID 000a1da7d9d7e35b: http://www.ebi.ac.uk/ena/data/view/ERS919915

If we no longer receive IDs matching this field in EBI data, our ability to retrieve data for American Gut will be broken.

antgonza:
Yes, it will be good to add a comment explaining those "magic" numbers.

squirrelo:
Those should be 1 since the AG human survey is 1, good catch. Fixed. The data going to open humans will be exactly the same, this just updates so the code grabs the correct survey ID for the main human survey.

PR #611, opened 2016-09-06
new unit test module for auth_handlers.py

sjanssen2:
My first try to add proper unit tests for a handler class. All comments are appreciated.

### DIFF
--- a/amgut/test/test_auth_handlers.py
+++ b/amgut/test/test_auth_handlers.py
@@ -83,10 +83,6 @@ class TestAuthLoginHandler(TestHandlerBase):
         port = self.get_http_port()
         self.assertEqual(response.effective_url, 'http://localhost:%d/' % port)
 
-        # check if url is wrong. SJ: Is this in general a good idea?
-        response = self.get('/authh/login/')
-        self.assertEqual(response.code, 404)
-
     def test_post_correct_pass(self):
         # test a sucessful login, for a handout kit
         self.mock_login('tst_ACJUJ')
@@ -114,10 +110,6 @@ class TestAuthLoginHandler(TestHandlerBase):
                                               'passwd': 'wrong'})
         self.assertIn(text_locale['handlers']['INVALID_KITID'], response.body)
         self.assertEqual(response.code, 200)
-
-    def test_set_current_user(self):
-        # TODO: add proper test for this once figure out how. Issue 567
-        pass
 
 
 class TestAuthLogoutHandler(TestHandlerBase):
--- a/amgut/test/test_auth_handlers.py
+++ b/amgut/test/test_auth_handlers.py
@@ -109,7 +109,7 @@ class TestAuthLoginHandler(TestHandlerBase):
         self.assertIn(text_locale['handlers']['INVALID_KITID'], response.body)
         self.assertEqual(response.code, 200)
 
-        # check that invalid Password is reported for wrong password
+        # check that invalid password is reported for wrong password
         response = self.post('/auth/login/', {'skid': 'tst_ACJUJ',
                                               'passwd': 'wrong'})
         self.assertIn(text_locale['handlers']['INVALID_KITID'], response.body)
--- a/amgut/test/test_auth_handlers.py
+++ b/amgut/test/test_auth_handlers.py
@@ -1,0 +1,133 @@
+from unittest import main
+
+from tornado.escape import url_escape
+
+from amgut.connections import ag_data
+from amgut.test.tornado_test_base import TestHandlerBase
+from amgut import text_locale
+
+
+class TestAuthBasehandler(TestHandlerBase):
+    def test_set_current_user(self):
+        # test if new cookie will be set
+        response = self.get('/auth/register/')
+        self.assertIn('Set-Cookie', response.headers.keys())
+
+        # test that cookie is re-used
+        self.mock_login('tst_ACJUJ')
+        response = self.get('/auth/register/')
+        self.assertNotIn('Set-Cookie', response.headers.keys())
+
+
+class TestAuthRegisterHandoutHandler(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/auth/register/')
+        self.assertEqual(response.code, 200)
+        port = self.get_http_port()
+        self.assertEqual(response.effective_url,
+                         'http://localhost:%d/auth/register/' % port)
+
+    def test_post(self):
+        # test reaction to invalid kit IDs
+        response = self.post('/auth/register/',
+                             {'kit_id': 'wrong_kit_id',
+                              'password': 'test',
+                              'email': 'e@mail',
+                              'email2': 'e@mail',
+                              'participantname': 'joe',
+                              'address': 'street',
+                              'city': 'metropolis',
+                              'state': 'XX',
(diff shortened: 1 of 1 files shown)


sjanssen2:
@wasade @ElDeveloper @josenavas could you please look into the Travis error. I have the feeling that it is not related to my unit tests.

wasade:
Weird... either a config entry changed or something in the BG text? Not
seeing it in changes (or I'm blind) yet master is not having this issue?

sjanssen2:
Master is having the same problem: https://github.com/biocore/american-gut-web/pull/612 (I only added an empty file). @EmbrietteH are you aware of any config changes?

EmbrietteH:
Not that I am aware of...

sjanssen2:
found the bug: its a ’ character in the SHIPPING_ADDRESS of amgut/lib/locale_data/british_gut.py line 86.

sjanssen2:
@wasade @josenavas @EmbrietteH now that Travis passes, I am happy to get some feedback about my first unit test for AG.

amgut/test/test_auth_handlers.py
```diff
@@ -0,0 +1,133 @@
...
+                              'zip': '12345',
+                              'country': 'Afghanistan'})
+        port = self.get_http_port()
+        self.assertEqual(
+            response.effective_url,
+            'http://localhost:%d/?loginerror=%s' %
+            (port, url_escape(text_locale['handlers']['INVALID_KITID'])))
+
+        unregistered_kits = ag_data.get_all_handout_kits()
+        if len(unregistered_kits) <= 0:
```
josenavas: if you are modifying the database, you should add a `tearDown` method in the test class that reverts those changes. If you do that, this error will never be raised, so you can remove this code.

amgut/test/test_auth_handlers.py
```diff
@@ -0,0 +1,133 @@
...
+
+class TestAuthLoginHandler(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/auth/login/')
+        self.assertEqual(response.code, 200)
+        # make sure redirect happened properly
+        port = self.get_http_port()
+        self.assertEqual(response.effective_url, 'http://localhost:%d/' % port)
+
+        # check if url is wrong. SJ: Is this in general a good idea?
```
josenavas: I don't think this test is needed, because is not actually testing the Handler that we wrote, but rather the entire "webserver.py" and a basic functionality of tornado. Safe to remove :-)

amgut/test/test_auth_handlers.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertIn(text_locale['handlers']['INVALID_KITID'], response.body)
+        self.assertEqual(response.code, 200)
+
+        # check that invalid password is reported for wrong password
+        response = self.post('/auth/login/', {'skid': 'tst_ACJUJ',
+                                              'passwd': 'wrong'})
+        self.assertIn(text_locale['handlers']['INVALID_KITID'], response.body)
+        self.assertEqual(response.code, 200)
+
+    def test_set_current_user(self):
```
josenavas: I think this can be removed? I think you have a test for set current user on the top of this file - or am I reading wrong?

amgut/test/test_auth_handlers.py
```diff
@@ -0,0 +1,133 @@
...
+        pass
+
+
+class TestAuthLogoutHandler(TestHandlerBase):
+    def test_get(self):
+        response = self.get('/auth/login/')
+        self.assertEqual(response.code, 200)
+        # make sure redirect happened properly
+        port = self.get_http_port()
+        self.assertEqual(response.effective_url, 'http://localhost:%d/' % port)
```
josenavas: Is there a header set some in place here to remove the cookie? I noticed that for checking the log in you're checking the `Set-Cookie` header, so I'm wondering if there is something similar in here. If not it's ok!

josenavas:
Thanks @sjanssen2 minor comments!

amgut/test/test_auth_handlers.py
sjanssen2: Honestly, I don't know how I can undo the changed. I have the feeling that there is no python method to undo this specific operation, thus I need to operate on the database myself or revert the complete database?! Any suggestions how I should proceed here?

wasade: @josenavas, in the CI environment, the database is reset on every build, and on a local system, this would require probably thousands of runs before its an issue, right?

josenavas: Not sure - but I'm more concerned if a change in the order of the tests is going to fire up a stochastic failure. If we don't foresee this issue, what we can do is open an issue, put a TODO comment on the code with the link to the issue and try to fix later, so we don't loose track of that. I don't know from the top of my head how to revert this changes in the database, as that is specific to the code being executed, but I was told that in this repository, instead of reseting the full database in each test, the local changes are reverted

sjanssen2: I dived into the SQL code. When a kit is registered to an user, the kit is inserted into the table ag_kit and removed from the table ag_handout_kits. During this process, we loose the information in column "created_on". Thus, I cannot revert this function.
To have the DB in the same state as before the test, I would need to write code which anticipates what is going to happen, store the information which is going to be lost somewhere else, perform the SQL operation which I want to test, and create a new function which takes the otherwise stored information to revert the tested SQL operation. This sounds not only quite complicated, but surely introduces potential mistakes + the need to keep code in sync. 
What shall I do??

josenavas: Thanks for taking a look @sjanssen2 !! I think for the time being you can create a new issue, and put a comment in the code: `#TODO: see issue #XXX` so we can keep track of it.

I think at some point we can do something like what we are doing on Qiita ATM, reset the DB on each class (rather than test), but I think this will need a longer discussion.

amgut/test/test_auth_handlers.py
sjanssen2: I am not familiar with tornados magic, but I doubt that cookie deletion happens in the according two lines of the class AuthLogoutHandler(AuthBasehandler):
        self.set_current_user()
        self.redirect(media_locale['SITEBASE'] + "/")

amgut/test/test_auth_handlers.py
sjanssen2: I think you are right. It is basically an artifact from the template I was using to write my code, i.e. it is qiita's auth_handler unit test.

amgut/test/test_auth_handlers.py
sjanssen2: agree. I removed this check.

amgut/test/test_auth_handlers.py
sjanssen2: I opened the issue #618

amgut/test/test_auth_handlers.py
sjanssen2: How can I test that the cookie has been properly removed? It's not necessary here, but in labadmin.

wasade: Not sure. SA may help. Alternatively, if the test is to check whether you're logged in, what about testing getting a 403 when accessing `/authed/portal`?

PR #613, opened 2016-09-06
bugfix: changed ' char in shipping address

sjanssen2:
Replacing a strange char in the shipping address of BritishGut, thus Trivis tests should no longer fail.

### DIFF
--- a/amgut/lib/locale_data/british_gut.py
+++ b/amgut/lib/locale_data/british_gut.py
@@ -83,7 +83,7 @@ media_locale = {
     'SWAB_HANDLING': "http://player.vimeo.com/video/62393487",
     'EMAIL_ERROR': "There was a problem sending your email. Please contact us directly at <a href='mailto:%(help_email)s'>%(help_email)s</a>" % {'help_email': HELP_EMAIL},
     'EMAIL_SENT': 'Your message has been sent. We will reply shortly',
-    'SHIPPING_ADDRESS': 'British Gut<br/>Department of Twin Research<br>Kings College London<br/>St Thomas’ Hospital Campus<br>The Rayne Institute<br>Lambeth Wing, 4th Floor<br/>Westminster Bridge Road<br>London SE1 7EH'
+    'SHIPPING_ADDRESS': 'British Gut<br/>Department of Twin Research<br>Kings College London<br/>St Thomas\' Hospital Campus<br>The Rayne Institute<br>Lambeth Wing, 4th Floor<br/>Westminster Bridge Road<br>London SE1 7EH'
 }
 
 _HANDLERS = {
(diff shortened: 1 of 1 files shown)


sjanssen2:
@EmbrietteH can you merge that directly, or do we need a "code review"?

josenavas:
Hi @sjanssen2, thanks for detecting the issue and fixing it! However, it looks like this was working before and then it stop working at some point - do we know why?
My concern is that the error was `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 117: ordinal not in range(128)`. That means that the character is not ascii, but we are supporting UTF-8 and that character is UTF-8:
```python
In [1]: s = 'homas’ Hospita'

In [2]: s.decode('ascii')
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-2-735804de5fd4> in <module>()
----> 1 s.decode('ascii')

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 5: ordinal not in range(128)

In [3]: s.decode('utf-8')
Out[3]: u'homas\u2019 Hospita'
```
We have a long story with problems with UTF-8. If the issue is that one of our dependencies changed the default behavior, we need to exactly troubleshoot the source of the problem and ensure that the code is UTF-8 compliant.

This change makes the tests pass, but this will make the real system to fail if there are UTF-8 characters in the address, which they'll be from past experience.

Can you take a closer look and see where the error is raised? It may be better to actually change the code so it decodes it using UTF-8 rather than change the test to make sure it is ASCII...

Again, thanks a lot for finding the issue and spending the time fixing it!

sjanssen2:
Phu, I just started to understand the code base. Since this issue maybe blocks other developments, I'd like to suggest that someone more experienced should work on hunting it down - might be way faster than letting me find out which dependency turned faulty over time.

wasade:
@josenavas, the full traceback is [here](https://travis-ci.org/biocore/american-gut-web/builds/157948309#L278) and shows that the error is on import by nose. It is possible that nose has changed versions as `setup.py` is nonspecific. It seems as though nose or python is disregarding the [encoding](https://github.com/biocore/american-gut-web/blob/master/amgut/lib/locale_data/british_gut.py#L2). I agree it is confusing, but right now the evidence suggests this is either with python itself or a test harness dependency. 

@sjanssen2, can prefix the string with `u` to explicitly denote the string as unicode and see if that works?

sjanssen2:
prefixing the string which holds the utf-8 char works!

josenavas:
:boom:

josenavas:
@wasade do you think that this needs further investigation or it should be safe for the rest of the code?

PR #619, opened 2016-09-20
New platemapper DB schema layout

josenavas:
@wasade @antgonza @qiyunzhu @sjanssen2 can you review?

I think this adds a good starting point that we can modify as specific details of the plate mapper raise.


josenavas:
For reference, here is the relevant part of the HTML:

screen shot 2016-09-20 at 10 14 58 am

wasade:
Nothing stood out to me on review but CI is complaining

josenavas:
Thanks @wasade - I sometimes hate DBschema - it looks like there are some name conflicts - fixing right now

josenavas:
💥

qiyunzhu:
That's a wonderful work in such a short time period! I have a few questions after reading your design:

My major question is that, to my understanding, properties such as `tm300_8_tool`, `master_mix_lot` and `water_lot` are NOT properties of specific target genes, but properties of specific protocols. Even if the wet lab are always sequencing 16S rRNA, they are likely replacing their TM300-8 tools from time to time.

The table name `plate_map` is ambiguous to me. Maybe use `sample_plate_sample` or `sample_plate_to_sample`?

I am thinking of replacing the ambiguous term `template` with `barcode_seq_plate`, which emphasizes the fact that this is a plate. Meanwhile, "template" may mean other things in the future, such as "protocol template"

Also, I am thinking about a mechanism that differentiates "simple" fields (such as `tm300_8_tool`) and "plate-like" fields (such as `template`, and perhaps some `primer_plate`s in the future). The plate-like fields will have same or similar structures, and they will be handled by the same user interface. Does that make sense?

Shall we pre-populate some fields with options, such as:
`INSERT INTO pm.plate_type (name, cols, rows, notes) VALUES ('96-well', 12, 8, 'Standard 96-well plate');`

Best,
Qiyun

qiyunzhu:
Edition: I think I misunderstood the original design. "protocol_target_gene" makes sense to me now.

josenavas:
I've addressed your comments @qiyunzhu 

I've change `plate_map` to `sample_plate_sample` as this will make it follow the other tables that are used to "join" the tables, in this case, this one is holding the relations between the table `sample_plate` and `sample`.
 
Changed `template` to `barcode_sequence_plate` and relevant column names.

Regarding simple vs plate like fields. Not sure how that will affect the interface, I'm unaware of deciding that at this point until we start playing with the system. I don't think that matters for the design of the DB.

Regarding pre-population of fields. I don't think it is blocking at this point (you don't need that for development) and I would like to get the full list of all values at once by the wet lab, so we add all the values at the end (I don't currently have all of them, maybe later today)

qiyunzhu:
@josenavas Thanks for your feedback! Make sense to me! I guess that it will helpful that we ask the wet lab and a get a full list of values. For now, we can already play with what we already have.

josenavas:
@wasade @antgonza if there are no more comments, can this be merged so @qiyunzhu and @sjanssen2 can start working on the python code and the interface?

PR #620, opened 2016-09-21
Fixing bugs discovered by Qiyun

josenavas:


qiyunzhu:
Hi Jose, I looked over your modification. Everything looks fine. When I mentioned "barcode_sequence_plate_layout", I actually meant that we may rename "sample_plate_sample" as "sample_plate_layout". But this does not matter much. I haven't seen any other issues so far. I will need bit time to figure out how to read / write time stamp in Python and in JavaScript, but everything else should be very straightforward. - Qiyun

josenavas:
Ah ok - let me rename that, I actually like it more.

Regarding the timestamp - when you query it you will receive a python's [datetime](https://docs.python.org/2/library/datetime.html) object so you don't need to do anything special on the python side. You can also easily [serialize](https://docs.python.org/2/library/datetime.html#datetime.date.strftime) it in any format that you want.

For insertion you can just call `now()` in the SQL (see an example [here](https://github.com/biocore/qiita/blob/5345a850e814df95d9629f79578d62a72c9561fc/qiita_db/logger.py#L100)).

josenavas:
@qiyunzhu I've addressed your comments - can you do another review? Thanks!

qiyunzhu:
@josenavas thank you for the explanation! The code now looks all good to me. I will follow the Python datetime tip you shared. Re +1: should I just type "+1" here?

josenavas:
Yeah, is you just type +1 is ok - or any comment just to say that you agree with the code. Usually we just type `:+1:` which gets transformed to 👍 :-D

@wasade can you take a quick look and merge? @qiyunzhu is working on the python code on labadmin so having this in would be useful. Thanks!

qiyunzhu:
Gotta! :D

PR #632, opened 2017-01-11
Bugfix surveyoverview

sjanssen2:
Fix for:
An error message is popping up when participants try to access the participant profile for their animal samples. We have tried on the test account as well and find the same error when hovering over Animal Samples --> clicking animal participant profile. Just wanted to give you a heads up!

### DIFF
--- a/amgut/templates/participant_overview.html
+++ b/amgut/templates/participant_overview.html
@@ -18,30 +18,53 @@
         <table class="list" id="survey" width="100%">
 {% if participant_type == 'human' %}
             <tr>
-                <td>{% raw tl['COMPLETED_CONSENT'] %}</td>
+                <td style="text-align: left">{% raw tl['COMPLETED_CONSENT'] %}:</td>
                 <td><img src="{% raw media_locale['SITEBASE'] %}/static/img/icon-green-checkmark.png"/></td>
             </tr>
 {% end %}
             <tr>
-                <td>{% raw tl['COMPLETED_SURVEY'] %}
-{% if participant_type == 'human' %}
-                    <form id="edit_survey" name="edit_survey" method="post" action="{% raw media_locale['SITEBASE'] %}/authed/human_survey/">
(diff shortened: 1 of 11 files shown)


sjanssen2:
@EmbrietteH @josenavas @wasade please review this PR which addresses the bug with the rendering of animal samples.

Problem was that all old animals where also present with the same participant name as humans (which doesn't make sense). Only lately added animals did not have human versions. Since rendering relied on human information, it failed for those newer animals.

sjanssen2:
I cannot add useful unit tests, since we don't have a DB with animal or secondary surveys. Thus, coverage decreases.

amgut/templates/participant_overview.html
```diff
@@ -18,30 +18,53 @@
...
-                    </form>
-                    {% raw vioscreen_text %}
-{% elif participant_type == 'animal' %}
-                    <form id="edit_survey" name="edit_survey" method="get" action="{% raw media_locale['SITEBASE'] %}/authed/add_animal/">
-                        <input type="hidden" name="survey" value={{ survey_id}}>
-                        <input type="button" name="submit_button" id="submit_button" value="edit" onclick="submitEdit()">
-                    </form>
+                <td style="text-align: left">{% raw tl['COMPLETED_SURVEY'] %}:
+					<ol>
+{% for survey in internal_surveys %}
```
wasade: would it be possible to perform a tuple expansion here to avoid the indexing and magic numbers?

wasade:
only had 1 minor comment

amgut/templates/participant_overview.html
sjanssen2: that was my first thought, but there are too many information that are different between those two survey types such that I decided to have a more speaking if condition rather than one more cryptic loop that covers both cases.
Is that OK with you?

sjanssen2:
ready to merge @wasade ?

EmbrietteH:
ping @josenavas ping @mortonjt

sjanssen2:
Thanks guys!

Issue #636, opened 2017-02-02
Cannot remove existing participant consents and surveys 


elainewolfe:
Clicking on "Remove {participant name}" generates an error message, and does not delete the participant from the account. The participant should be deleted when clicking the remove button, and assigned samples (if any) are automatically unassigned.

sjanssen2:
I cannot replicate this with an arbitrary kit on my installation of labadmin and american-gut-web repositories :-/

elainewolfe:
To elaborate on the issue, if a participant does have samples assigned, clicking on Remove will unassign the barcodes, but will not remove the consent/surveys. This is on the live system.

sjanssen2:
Our test server is up again, but I cannot reproduce the error there either :-/

sjanssen2:
Could you please mail (sjanssen@ucsd.edu) me the affected kitID and the participants that should be removed. Please don't post those infos here, but use email.

sjanssen2:
Thanks for your mail, it helped since I could now get a more speaking error message:
Error: update or delete on table "ag_consent" violates foreign key constraint "fk_ag_login_surveys0" on table "ag_login_surveys"

sjanssen2:
@josenavas @antgonza do you guys know if SQL transactions are implemented for american gut web?

sjanssen2:
I think the problem is that american-gut-web/amgut/lib/data_access deleteAGParticipantSurvey expects to have only one survey for each participant. The one you sent me has two and thus not all entries are correctly removed from the according db tables.

sjanssen2:
My bugfix has been merged. We are now waiting that someone deploys it to the life system.

EmbrietteH:
Thanks Stefan, we'll deploy to the test system first and test it there to
make sure nothing weird was introduced, then we'll deploy to the live
system. Jose can deploy to test.

Embriette R. Hyde, PhD

Project Manager, American Gut Project
Assistant Project Scientist
Rob Knight Lab
Department of Pediatrics
University of California, San Diego
Phone: 832-845-7532

josenavas:
Just updated the test system - please make sure that everything behaves as expected.

EmbrietteH:
Confirmed that I can remove a participant; however, this does not work if there is a sample already assigned. If a sample is assigned, the participant should be able to click "Remove NAME," which will trigger un-assignment of the sample and removal of the participant.

EmbrietteH:
[...]

PR #637, opened 2017-02-03
fixed false assumption that a participant (source) has only one surve…

sjanssen2:
…y_id, when deleting this participant.
See issue #636

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -315,7 +315,7 @@ class AGDataAccess(object):
             TRN.add(sql, [tuple(survey_ids)])
 
             sql = "DELETE FROM survey_answers_other WHERE survey_id IN %s"
-            TRN.add(sql, [tuples(survey_ids)])
+            TRN.add(sql, [tuple(survey_ids)])
 
             # Reset survey attached to barcode(s)
             for info in self.getParticipantSamples(ag_login_id,
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -311,31 +311,26 @@ class AGDataAccess(object):
                 survey_ids.add(hit[0])
                 participant_emails.add(hit[1])
 
-            sql = "DELETE FROM survey_answers WHERE survey_id = %s"
-            for survey_id in survey_ids:
-                TRN.add(sql, [survey_id])
+            sql = "DELETE FROM survey_answers WHERE survey_id IN %s"
+            TRN.add(sql, [tubple(survey_ids)])
 
-            sql = "DELETE FROM survey_answers_other WHERE survey_id = %s"
-            for survey_id in survey_ids:
-                TRN.add(sql, [survey_id])
+            sql = "DELETE FROM survey_answers_other WHERE survey_id IN %s"
+            TRN.add(sql, [tuples(survey_ids)])
 
             # Reset survey attached to barcode(s)
             for info in self.getParticipantSamples(ag_login_id,
                                                    participant_name):
                 self.deleteSample(info['barcode'], ag_login_id)
 
-            sql = "DELETE FROM promoted_survey_ids WHERE survey_id = %s"
-            for survey_id in survey_ids:
-                TRN.add(sql, [survey_id])
+            sql = "DELETE FROM promoted_survey_ids WHERE survey_id IN %s"
+            TRN.add(sql, [tuple(survey_ids)])
 
             # Delete last due to foreign keys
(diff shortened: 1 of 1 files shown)


sjanssen2:
@EmbrietteH @wasade @josenavas please have a look at this PR which should fix the issue #636

sjanssen2:
Please think about fact that a participant might have multiple email addresses. If so, do we want to list them only once in the ag.consent_revoked table or multiple times each with a different email?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -303,28 +303,39 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
...
+            # former might be more than one.
+            survey_ids = set()
+            participant_emails = set()
+            for hit in TRN.execute_fetchindex():
+                survey_ids.add(hit[0])
+                participant_emails.add(hit[1])
 
             sql = "DELETE FROM survey_answers WHERE survey_id = %s"
-            TRN.add(sql, [survey_id])
+            for survey_id in survey_ids:
```
josenavas: Change to:
```python
sql = "DELETE FROM survey_answers WHERE survey_id IN %s"
TRN.add(sql, [tuple(survey_ids)])
```
Note the call to tuple is required. This way all the delete is done in a single SQL query.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -303,28 +303,39 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
...
+                TRN.add(sql, [survey_id])
 
             sql = "DELETE FROM survey_answers_other WHERE survey_id = %s"
-            TRN.add(sql, [survey_id])
+            for survey_id in survey_ids:
```
josenavas: Same as my previous comments - modify to use a single SQL query.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -303,28 +303,39 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
...
+                TRN.add(sql, [survey_id])
 
             # Reset survey attached to barcode(s)
             for info in self.getParticipantSamples(ag_login_id,
                                                    participant_name):
                 self.deleteSample(info['barcode'], ag_login_id)
 
             sql = "DELETE FROM promoted_survey_ids WHERE survey_id = %s"
-            TRN.add(sql, [survey_id])
+            for survey_id in survey_ids:
```
josenavas: Same here and in the next 3 for loops.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -333,7 +344,9 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
             sql = """INSERT INTO ag.consent_revoked
                      (ag_login_id,participant_name, participant_email)
                      VALUES (%s, %s, %s)"""
-            TRN.add(sql, [ag_login_id, participant_name, participant_email])
+            for participant_email in list(participant_emails):
```
josenavas: I don't think the call to `list` is needed (you can iterate over a set). However, this can be changed to:
```python
sql_args = [[ag_login_id, participant_name, pemail] for pemail in participant_emails]
TRN.add(sql, sql_args, many=True)
```

And add at the end of the function:
```python
TRN.execute()
```

So we make sure that whatever is listed here is actually executed before returning this functions (but not necessarily committed)

sjanssen2:
test are not possible, because there is no (survey_login_id, participant_name) tuple that appears more than once :-(

wasade:
:+1: following @josenavas's comments

amgut/lib/data_access/ag_data_access.py
sjanssen2: I like the tuple style. And will change the code to it.
A question for understanding: I thought that all SQL statements of this function are executed as a transaction, so why do I have to explicitly care here to trigger only one SQL query?

josenavas: Even it is a single transaction the number of sql queries changes and each sql queries has some overhead. Reducing the number of SQL queries reduces the running time of the method.

josenavas: Aaand clearly I need more coffee after reading my english on my last comment 😄

sjanssen2: Understood :-) It's no issue of concurrence, but performance. Good to know.

qiyunzhu:
I sat with @sjanssen2 and read the PR. It looks all good to the best of my knowledge.

PR #638, opened 2017-02-06
Adding scrub script

josenavas:
I've updated the DB schema and HTML files to match the contents of the live database. The only tables missing where the labadmin user tables.

I'm also adding to the repository a script to scrub the database for identifiable information. Please, review that file carefully and raise any concern and/or missing columns/tables that I've missed.

I'm currently running the script in a screen in the test system with a copy of last night's database. It is taking quite a while to run, which I think it is normal due to the nature of the operations and the size of the DB.

### DIFF
--- a/amgut/db/scrub.sql
+++ b/amgut/db/scrub.sql
@@ -32,6 +32,21 @@ CREATE FUNCTION random_string(length integer, source varchar, suffix varchar) RE
     END;
 $$ language plpgsql;
 
+-- Function to generate a random latitude or longitude
+CREATE FUNCTION random_lat_or_long() RETURNS varchar AS $$
+    DECLARE
+        result varchar;
+        numsource varchar;
+    BEGIN
+        numsource := '0123456789';
+        result := random_string(2, numsource, '') || ',' || random_string(5, numsource, '');
+        IF random() < 0.5 THEN
+            result := '-' || result;
+        END IF;
+        RETURN result;
+    END;
+$$ language plpgsql;
+
 -- Function to generate random emails
 CREATE FUNCTION random_email(column_name varchar, table_name varchar) RETURNS varchar AS $$
     DECLARE
@@ -52,6 +67,10 @@ CREATE FUNCTION random_date() RETURNS varchar AS $$
         RETURN result;
     END
 $$ language plpgsql;
+
+-- Create a temp table for storing the zipcodes
+CREATE TEMP TABLE allzipcodes ON COMMIT DROP AS
+    SELECT DISTINCT zip_code FROM ag.ag_human_survey;
 
 -- We have all the functions that we need, start scrubbing data
 DO $do$
@@ -176,7 +195,7 @@ BEGIN
                 parent_2_name = random_string(10, source3, 'Name - '),
                 birth_date = random_date(),
                 phone_num = random_string(10, numsource, ''),
-                zip_code = random_string(5, numsource, ''),
+                zip_code = (SELECT zip_code FROM allzipcodes ORDER BY RANDOM() LIMIT 1),
                 foodallergies_other_text = random_string(50, source4, 'Free text - '),
                 race_other = random_string(20, source5, 'Free text - '),
                 antibiotic_condition = random_string(30, source6, 'Free text - '),
@@ -252,8 +271,8 @@ BEGIN
                 city = random_string(10, source2, 'City - '),
                 state = random_string(10, source3, 'State - '),
                 zip = random_string(5, numsource, ''),
-                latitude = random_string(2, numsource, '') || ',' || random_string(5, numsource, ''),
-                longitude = random_string(2, numsource, '') || ',' || random_string(5, numsource, '')
+                latitude = random_lat_or_long(),
+                longitude = random_lat_or_long()
             WHERE ag_login_id = rec.ag_login_id;
     END LOOP;
 
@@ -299,14 +318,13 @@ BEGIN
 
     -- Table: ag_kit; columns: kit_password, kit_verification_code, open_humans_token
     source := retrieve_chars('kit_verification_code', 'ag.ag_kit');
-    source2 := retrieve_chars('open_humans_token', 'ag.ag_kit');
     FOR rec IN
         SELECT ag_kit_id FROM ag.ag_kit
     LOOP
         UPDATE ag.ag_kit
             SET kit_password = passwd,
                 kit_verification_code = random_string(5, source, ''),
-                open_humans_token = random_string(10, source2, '')
+                open_humans_token = NULL
             WHERE ag_kit_id = rec.ag_kit_id;
     END LOOP;
 
@@ -318,7 +336,7 @@ BEGIN
     LOOP
         UPDATE ag.ag_kit_barcodes
             SET notes = random_string(50, source, 'Free text - '),
-                other_text = random_string(50, source, 'Free text - ')
+                other_text = random_string(50, source2, 'Free text - ')
             WHERE ag_kit_barcode_id = rec.ag_kit_barcode_id;
     END LOOP;
 
@@ -341,3 +359,4 @@ DROP FUNCTION retrieve_chars(varchar, varchar);
 DROP FUNCTION random_string(integer, varchar, varchar);
 DROP FUNCTION random_email(varchar, varchar);
 DROP FUNCTION random_date();
+DROP FUNCTION random_lat_or_long();
--- a/amgut/db/scrub.sql
+++ b/amgut/db/scrub.sql
@@ -87,7 +87,7 @@ BEGIN
     numsteps := 18;
     currstep := 1;
     -- To simplify testing, all the passwords are going to be the same ('test')
-    passwd := '$2a$12$rX8UTcDkIj8bwcxZ22iRpebAxblEclT83xBiUIdJGUJGoUfznu1RK';
+    passwd := '$2a$10$2.6Y9HmBqUFmSvKCjWmBte70WF.zd3h4VqbhLMQK1xP67Aj3rei86';
     numsource := '0123456789';
 
     RAISE NOTICE 'STEP % of %: Gathering all characters', currstep, numsteps;
--- a/amgut/db/scrub.sql
+++ b/amgut/db/scrub.sql
@@ -33,27 +33,23 @@ CREATE FUNCTION random_string(length integer, source varchar, suffix varchar) RE
 $$ language plpgsql;
 
 -- Function to generate a random latitude or longitude
-CREATE FUNCTION random_lat_or_long() RETURNS varchar AS $$
+CREATE FUNCTION random_lat_or_long() RETURNS float8 AS $$
     DECLARE
         result varchar;
         numsource varchar;
     BEGIN
         numsource := '0123456789';
-        result := random_string(2, numsource, '') || ',' || random_string(5, numsource, '');
+        result := random_string(2, numsource, '') || '.' || random_string(5, numsource, '');
         IF random() < 0.5 THEN
             result := '-' || result;
         END IF;
-        RETURN result;
+        RETURN result::float8;
     END;
 $$ language plpgsql;
 
 -- Function to generate random emails
-CREATE FUNCTION random_email(column_name varchar, table_name varchar) RETURNS varchar AS $$
-    DECLARE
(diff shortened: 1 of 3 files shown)


josenavas:
Forgot to mention that the scrub script should be self documented, indicating which table/columns I'm scrubbing.

EmbrietteH:
This is a high priority PR, please review promptly!

EmbrietteH:
ping @wasade @sjanssen2 @mortonjt @tkosciol @ElDeveloper

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+
+    --Table: ag_kit_barcodes; columns: notes, other_text
+    source := retrieve_chars('notes', 'ag.ag_kit_barcodes');
+    source2 := retrieve_chars('other_text', 'ag.ag_kit_barcodes');
+    FOR rec IN
+        SELECT ag_kit_barcode_id FROM ag.ag_kit_barcodes
+    LOOP
+        UPDATE ag.ag_kit_barcodes
+            SET notes = random_string(50, source, 'Free text - '),
+                other_text = random_string(50, source, 'Free text - ')
```
wasade: source2

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
+-- This script is used to scrub the live American Gut database in order to
+-- generate a test database for Travis without personal information but
+-- representative of the real world.
+
+-- Function to generate a string with all the different characters present in
+-- a column (adapted from: http://dba.stackexchange.com/a/83679)
+CREATE FUNCTION retrieve_chars(column_name varchar, table_name varchar) RETURNS varchar AS $$
```
wasade: I'd feel substantially better about this approach if it the result was always assured to include `[a-z]` to handle possible low entropy responses.

Given that the goal is to ensure proper handling of non-ascii characters, is this method suitable and tested to be suitable for non-ascii? The SE response says so, but given the nature of this PR, it would be good to have added confidence. 

I'm not sure to the extent we need to be concerned, but UTF-8 is not the only character set we encounter, so it may be worth asserting separately that this method can handle other common ones like UTF-16, ISO-8859-1, and ANSI

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+        RETURN random_string(10, source, '') || '@' || random_string(5, source, '') || '.' || random_string(3, source, '');
+    END;
+$$ language plpgsql;
+
+-- Function to generate random dates
+CREATE FUNCTION random_date() RETURNS varchar AS $$
+    DECLARE
+        result varchar;
+    BEGIN
+        SELECT date(now() - trunc(random()  * 80) * '1 year'::interval - trunc(random() * 12) * '1 month'::interval - trunc(random() * 30) * '1 day'::interval) INTO result;
```
wasade: This will occasionally produce invalid dates

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+    FOR rec IN
+        SELECT ag_login_id, participant_name FROM ag.ag_human_survey
+    LOOP
+        UPDATE ag.ag_human_survey
+            SET participant_name = random_string(10, source, 'Name - '),
+                parent_1_name = random_string(10, source2, 'Name - '),
+                parent_2_name = random_string(10, source3, 'Name - '),
+                birth_date = random_date(),
+                phone_num = random_string(10, numsource, ''),
+                zip_code = random_string(5, numsource, ''),
```
wasade: Not all of the zipcodes are 5 chars, and not all are numeric

amgut/db/scrub.sql
wasade: Oh, I see, this is over all rows. [a-z] probably not necessary, but asserting independently that this method handles non-utf8 as expected is still a good idea.

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+    source3 := retrieve_chars('state', 'ag.ag_login');
+    FOR rec IN
+        SELECT ag_login_id FROM ag.ag_login
+    LOOP
+        UPDATE ag.ag_login
+            SET email = random_email('email', 'ag.ag_login'),
+                name = random_string('name', 'ag.ag_login'),
+                address = random_string(4, numsource, '') || random_string(10, source, ' '),
+                city = random_string(10, source2, 'City - '),
+                state = random_string(10, source3, 'State - '),
```
wasade: The state is normally stored as a two letter abbreviation, right?

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+                zip = random_string(5, numsource, ''),
+                latitude = random_string(2, numsource, '') || ',' || random_string(5, numsource, ''),
```
wasade: Negative values for both lat/long?

amgut/db/scrub.sql
```diff
@@ -0,0 +1,343 @@
...
+    -- Table: ag_kit; columns: kit_password, kit_verification_code, open_humans_token
+    source := retrieve_chars('kit_verification_code', 'ag.ag_kit');
+    source2 := retrieve_chars('open_humans_token', 'ag.ag_kit');
+    FOR rec IN
+        SELECT ag_kit_id FROM ag.ag_kit
+    LOOP
+        UPDATE ag.ag_kit
+            SET kit_password = passwd,
+                kit_verification_code = random_string(5, source, ''),
+                open_humans_token = random_string(10, source2, '')
```
wasade: why not just delete it since we're already invalidating it?

amgut/db/scrub.sql
josenavas: The database should be configured as UTF8 so non-utf8 characters will fail to be inserted in the DB directly, right?

amgut/db/scrub.sql
josenavas: Can you elaborate?

amgut/db/scrub.sql
josenavas: Gotcha - changes to grab a random zipcode from the table itself.

amgut/db/scrub.sql
josenavas: Actually, looking at the information in the DB it looks like it even contains some kind of free text...

amgut/db/scrub.sql
josenavas: Good call, modified

amgut/db/scrub.sql
josenavas: Sounds good!

amgut/db/scrub.sql
josenavas: Good call

amgut/db/scrub.sql
wasade: Not sure. It's the same server as qiita though

amgut/db/scrub.sql
wasade: february

amgut/db/scrub.sql
wasade: Want to note that as a point of normalization? Not sure how that would happen so wonder if it's from the original db

josenavas:
For reference, [here](https://github.com/biocore/american-gut-web/files/764329/scrub.txt) is the old scrub script (note that I renamed to txt so github allows me to attach it here).

@wasade how does a metadata pulldown makes a difference in here given that the metadata pulldown is already not including PHI information?:

josenavas:
BTW, forgot to mention that I'm currently running this in the test system and will set it up with a test server once it is done running.

amgut/db/scrub.sql
josenavas: It is working with intervals so it is actually generating correct dates:
```sql
jose=# SELECT date(now() + 19 * '1 day'::interval);
    date    
------------
 2017-02-28
(1 row)

jose=# SELECT date(now() + 20 * '1 day'::interval);
    date    
------------
 2017-03-01
(1 row)
```

amgut/db/scrub.sql
josenavas: Just to see if I understand you correctly, are you suggesting opening an issue?

wasade:
It is in effect an integration test to assert that the database is still coherent. The per sample metadata results should be the same with the exception of the HSIs, however those with multiple samples should still have a common HSI, right?

josenavas:
Yup that makes sense.

I just noticed that the scrub script is still running (has been running for more than 2 hours), I think this is taking too long? Probably is due to all the random values that we are generating... ugh

wasade:
Ya, I don't think it's the size as the db is relatively small but definitely quite a few numbers are necessary

josenavas:
Ok, the script is still running since yesterday... I guess this is not the best solution?

@wasade do you have any suggestions on how we can rewrite the script in a more efficient way?

sjanssen2:
I bet the most time is spend with the "shuffling" of the data. 
I saw that you can obtain a Pandas dataframe from a SQL query. Can't we do the shuffling within pandas instead of SQL. The question is: how much time does it take to write back from Pandas to SQL?

josenavas:
Agree that is probably the source. I can try to write an alternative in python, but this will fly only if the python shuffle is way faster than the SQL shuffle.

josenavas:
@wasade @sjanssen2 The time sink was not actually the randomization of the strings, but actually the number of for loops in the code - apparently for loops in SQL are extremely slow. Also, gathering the different characters was expensive and it was done multiple times in different spots.

I've taken a different approach in which I first grab all the different characters and then randomize the values in the different columns. As seen below, this takes now ~10 min to run, which I think it is acceptable.

```
[ag_dev@ag-test ~]$ date; psql -U postgres -d agjosetest -f scrub.sql ; date
Mon Mar  6 09:10:30 PST 2017
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
CREATE FUNCTION
SELECT 2081
CREATE TABLE
psql:scrub.sql:295: NOTICE:  STEP 1 of 18: Gathering all characters
psql:scrub.sql:295: NOTICE:  STEP 2 of 18: Gathering all email characters
psql:scrub.sql:295: NOTICE:  STEP 3 of 18: Scrubbing ag_survey_multiples_backup
psql:scrub.sql:295: NOTICE:  STEP 4 of 18: Scrubbing ag.ag_handout_kits
psql:scrub.sql:295: NOTICE:  STEP 5 of 18: Scrubbing barcodes.project
psql:scrub.sql:295: NOTICE:  STEP 6 of 18: Scrubbing ag.consent_revoked
psql:scrub.sql:295: NOTICE:  STEP 7 of 18: Scrubbing ag.ag_consent and ag.ag_login_surveys
psql:scrub.sql:295: NOTICE:  STEP 8 of 18: Scrubbing ag.ag_human_survey
psql:scrub.sql:295: NOTICE:  STEP 9 of 18: Scrubbing ag.survey_answers
psql:scrub.sql:295: NOTICE:  STEP 10 of 18: Scrubbing ag.survey_answers_other
psql:scrub.sql:295: NOTICE:  STEP 11 of 18: Scrubbing ag.ag_participant_exceptions
psql:scrub.sql:295: NOTICE:  STEP 12 of 18: Scrubbing ag.ag_animal_survey
psql:scrub.sql:295: NOTICE:  STEP 13 of 18: Scrubbing ag.ag_login
psql:scrub.sql:295: NOTICE:  STEP 14 of 18: Scrubbing ag.ag_survey_answer
psql:scrub.sql:295: NOTICE:  STEP 15 of 18: Scrubbing ag.ag_survey_multiples
psql:scrub.sql:295: NOTICE:  STEP 16 of 18: Scrubbing ag.ag_kit
psql:scrub.sql:295: NOTICE:  STEP 17 of 18: Scrubbing ag.ag_kit_barcodes
psql:scrub.sql:295: NOTICE:  STEP 18 of 18: Scrubbing labadmin user tables
DO
UPDATE 1
DROP FUNCTION
DROP FUNCTION
DROP FUNCTION
DROP FUNCTION
DROP FUNCTION
DROP TABLE
DROP TABLE
Mon Mar  6 09:20:57 PST 2017
```

wasade:
Just to check, has the website been run off the scrubbed result?

josenavas:
The test environment is currently working under the scrubbed result.

EmbrietteH:
Can this be merged?

PR #643, opened 2017-03-17
Prepare unit tests for DB update

sjanssen2:
Hey guys, I know this is a length one, but it didn't make sense to split.

In order to make all the cool things happen we discussed in today's meeting, we need to be able to write proper unit tests - something that has not been done to heavily for AG / labadmin, unfortunately.
Jose provided a refactored scrubbing script to remove sensitive information from the productive DB. This version will provide a lot of test data which we missed until now. To make the DB change happen we need to do three things:
1) remove some current assumptions about DB entries, like all kit_ids start in the current version with "tst_" but not in the update
2) do the same for labadmin code
3) actually switch DB dump

This PR addresses 1) and yes, I already checked that unittests work on both DB versions. I cannot test this on Travis right now.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -1152,13 +1152,13 @@ class AGDataAccess(object):
         Returns
         -------
         [[str]]
-            Example: [["Name - z\xc3\x96DOZ8(Z~'"],
-                      ["Name - z\xc3\x96DOZ8(Z~'"],
-                      ['Name - QpeY\xc3\xb8u#0\xc3\xa5<'],
-                      ['Name - S)#@G]xOdL'],
-                      ['Name - Y5"^&sGQiW'],
-                      ['Name - L\xc3\xa7+c\r\xc3\xa5?\r\xc2\xbf!'],
-                      ['Name - (~|w:S\xc3\x85#L\xc3\x84']]
+            Example: ["Name - z\xc3\x96DOZ8(Z~'",
+                      "Name - z\xc3\x96DOZ8(Z~'",
+                      'Name - QpeY\xc3\xb8u#0\xc3\xa5<',
+                      'Name - S)#@G]xOdL',
+                      'Name - Y5"^&sGQiW',
+                      'Name - L\xc3\xa7+c\r\xc3\xa5?\r\xc2\xbf!',
+                      'Name - (~|w:S\xc3\x85#L\xc3\x84']
 
         Raises
         ------
@@ -1174,7 +1174,7 @@ class AGDataAccess(object):
             if not info:
                 raise ValueError('ag_login_id not in database: %s' %
                                  ag_login_id)
-            return info
+            return [n[0] for n in info]
 
     def get_barcode_from_ag_login_id(self, ag_login_id):
         """ Returns all barcodes for a given ag_login_id.
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -1102,3 +1102,27 @@ class AGDataAccess(object):
                 raise ValueError('barcode not in database: %s' %
                                  ag_login_id)
             return [dict(row) for row in info]
+
+    def get_random_supplied_kit_id_unverified(self):
+        with TRN:
+            sql = """SELECT supplied_kit_id
+                     FROM ag.ag_kit
+                     WHERE ag.ag_kit.kit_verified = 'n'
+                     LIMIT 1"""
+            TRN.add(sql, [])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('No unverified kits in DB')
+            return info[0][0]
+
+    def get_ag_login_id_from_barcode(self, barcode):
+        with TRN:
+            sql = """SELECT ag.ag_kit.ag_login_id
+                     FROM ag.ag_kit_barcodes
+                     JOIN ag.ag_kit USING (ag_kit_id)
+                     WHERE ag.ag_kit_barcodes.barcode = %s"""
+            TRN.add(sql, [barcode])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('Barcode "%s" not in DB' % barcode)
+            return info[0][0]
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -970,7 +970,23 @@ class AGDataAccess(object):
             TRN.add(sql, [barcode])
             return TRN.execute_fetchlast()
 
+    # following are DB access functions only used for unit testing:
+
     def get_random_supplied_kit_id_scanned_unconsented(self):
+        """ Returns randomly chosen supplied_kit_id and barcode which has been
+        scanned but is without consent.
+        For unit testing only!
+
+        Returns
+        -------
+        str: [supplied_kit_id, barcode]
+            example: ['fNIYa', '000001053']
+
+        Raises
+        ------
+        ValueError
+            If no kits can be found in the DB that have been scanned and
+            are without consent."""
         with TRN:
             sql = """SELECT supplied_kit_id, barcode
                      FROM barcodes.barcode
@@ -986,19 +1002,19 @@ class AGDataAccess(object):
             return info[0]
 
     def get_random_handout_printed_min6_supplied_kit_id(self):
-        """ For testing: get a random supplied_kit_id with printed results
-            and 6 swaps per kit.
+        """ Returns a random supplied_kit_id with printed results
+        and 6 swaps per kit.
+        For unit testing only!
 
         Returns
         -------
         supplied_kit_id : str
-            A supplied_kit_id.
+            A supplied_kit_id. Example: 'DS_ubdvq'
 
         Raises
         ------
         ValueError
-            If no hand out kit exists, satisfing the given conditions.
-        """
+            If no hand out kit exists, satisfing the given conditions."""
         with TRN:
             sql = """SELECT kit_id
                      FROM ag.ag_handout_kits
@@ -1010,6 +1026,18 @@ class AGDataAccess(object):
             return info[0][0]
 
     def get_random_email(self):
+        """ Return randomly chosen email.
+        For unit testing only!
+
+        Returns
+        -------
+        str: email
+            Example: 'a03E9u6ZAu@glA+)./Vn'
+
+        Raises
+        ------
+        ValueError
+            If no emails be found in the DB."""
         with TRN:
             sql = """SELECT email
                      FROM ag.ag_login
@@ -1021,6 +1049,23 @@ class AGDataAccess(object):
             return info[0][0]
 
     def get_random_barcode(self, deposited=True):
+        """ Returns randomly chosen barcode.
+        For unit testing only!
+
+        Parameters
+        ----------
+        deposited : boolean
+            If true, pick a deposited barcode. Default = True
+
+        Returns
+        -------
+        str: barcode
+            Example: '000032951'
+
+        Raises
+        ------
+        ValueError
+            If no barcodes can be found in the DB."""
         depos = 'TRUE'
         if deposited is False:
             depos = 'FALSE'
@@ -1036,6 +1081,24 @@ class AGDataAccess(object):
             return info[0][0]
 
     def get_email_from_ag_login_id(self, ag_login_id):
+        """ Returns email for a given ag_login_id.
+        For unit testing only!
+
+        Parameters
+        ----------
+        ag_login_id : str
+            Existing ag_login_id.
+
+        Returns
+        -------
+        str: email
+            Example: 'xX/tEv7O+T@6Ri7C.)LO'
+
+        Raises
+        ------
+        ValueError
+            If ag_login_id is not in DB.
+        """
         with TRN:
             sql = """SELECT email
                      FROM ag.ag_login
@@ -1047,25 +1110,24 @@ class AGDataAccess(object):
             return info[0][0]
 
     def get_supplied_kit_id(self, ag_login_id):
-        """ Get supplied_kit_id for ag_login_id.
-
-        Useful for unit tests, since the ag_login_id is stable along different
-        database versions, while the supplied_kit_id is subject to scrubbing.
+        """ Returns supplied_kit_id for a given ag_login_id.
+        For unit testing only!
 
         Parameters
         ----------
         ag_login_id : str
-            A valid login ID, that should be a test as a valid UUID
+            Existing ag_login_id.
 
         Returns
         -------
-        supplied_kit_id : str
-            The supplied_kit_id for the given ag_login_id
+        str
+            The supplied_kit_id for the given ag_login_id.
+            Example: 'DokBF'
 
         Raises
         ------
         ValueError
-            Unknown ag_login_id passed
+            If ag_login_id is not in DB.
         """
         with TRN:
             sql = """SELECT supplied_kit_id
@@ -1079,6 +1141,30 @@ class AGDataAccess(object):
             return info[0][0]
 
     def get_participant_names_from_ag_login_id(self, ag_login_id):
+        """ Returns all participant_name(s) for a given ag_login_id.
+        For unit testing only!
+
+        Parameters
+        ----------
+        ag_login_id : str
+            Existing ag_login_id.
+
+        Returns
+        -------
+        [[str]]
+            Example: [["Name - z\xc3\x96DOZ8(Z~'"],
+                      ["Name - z\xc3\x96DOZ8(Z~'"],
+                      ['Name - QpeY\xc3\xb8u#0\xc3\xa5<'],
+                      ['Name - S)#@G]xOdL'],
+                      ['Name - Y5"^&sGQiW'],
+                      ['Name - L\xc3\xa7+c\r\xc3\xa5?\r\xc2\xbf!'],
+                      ['Name - (~|w:S\xc3\x85#L\xc3\x84']]
+
+        Raises
+        ------
+        ValueError
+            If ag_login_id is not in DB.
+        """
         with TRN:
             sql = """SELECT participant_name
                      FROM ag.ag_login_surveys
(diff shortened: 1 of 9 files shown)


sjanssen2:
Two things: @josenavas 
1) please ensure that your scrubbing script re-sets all passwords to "test". Otherwise, some unittests will fail and it will prohibit proper future debugging if we cannot login as an arbitrary kit.
2) we need to find a way to create ROLE ag_dev

sjanssen2:
I think I found a bug in amgut/lib/data_access/survey.py function fetch_survey():
I added the restriction that only those answers should be enumerated, that belong to the correct "survey_group". Tests are passing, but since our coverage is low I'd like you to ask to run some manual inspections.

sjanssen2:
Don't merge for know. I found that test results depend on execution order. I need to fix that first!

josenavas:
Thanks a lot @sjanssen2 

I think the best way to move forward is to get this PR complete until the point that we are sure that the DB is correct. We then do a PR merging the new DB, ask travis to re-execute the tests in here and merge, so everything is clean.

Thanks for the effors @sjanssen2 !

sjanssen2:
Hey @mortonjt @wasade @josenavas @EmbrietteH @jwdebelius I could now need one of those rapid reviews Rob was writing about in his mail :-)

josenavas:
Ok - starting now! I thought you needed me to fix the pulldown to finish this.

@mortonjt @wasade @EmbrietteH @jwdebelius Please do not merge this although reviews are welcome :)

mortonjt:
I'm seeing that there are a whole bunch of methods that are declared, but not tested or used.  I'm assuming that there will be more follow up PRs right?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
             sql = "SELECT deposited FROM ag.ag_kit_barcodes WHERE barcode = %s"
             TRN.add(sql, [barcode])
             return TRN.execute_fetchlast()
+
+    # following are DB access functions only used for unit testing:
+
+    def get_random_supplied_kit_id_scanned_unconsented(self):
```
mortonjt: What do you think about moving these testing functions to a separate file (i.e. `util_testing.py`)
It'll help future developers figure out where the testing functions are.

amgut/lib/data_access/test/test_survey.py
```diff
@@ -230,6 +229,56 @@ def test_american_name(self):
 
 
 class TestSurvey(TestCase):
+    def insert_data(self):
+        """ Prepare tests by inserting some data into the DB.
+
+        Returns
+        -------
+        survey object, survey_id, notes_test and consent of the newly created
```
mortonjt: Are these all strings?

amgut/lib/data_access/ag_data_access.py
sjanssen2: that's a great idea. Right now, I don't know how to do that such that still only one DB object is accessing the DB. It would also be help to re-use those functions in labadmin.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
...
+                     FROM ag.ag_login
+                     LIMIT 1"""
+            TRN.add(sql, [])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('No emails found.')
+            return info[0][0]
+
+    def get_random_barcode(self, deposited=True):
+        """ Returns randomly chosen barcode.
```
mortonjt: Maybe I'm missing something - how is this random?  And do we really want random tests?

sjanssen2: maybe random is not the right word. It returns the first hit in the DB (by using limit 1 in the SQL statement), thus it depends on the internal indexing of the DB which value is returned. Let's say it is quasi random ;-)

amgut/lib/data_access/ag_data_access.py
mortonjt: We might be able to subclass it - but perhaps that's a bit outside the scope of this PR

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -312,139 +318,137 @@ def test_getAnimalParticipantsNotPresent(self):
...
+        names = self.ag_data.get_participant_names_from_ag_login_id(i)
+        # collect results for ALL names
+        res = [self.ag_data.getParticipantSamples(i, name)
+               for name in names]
+        exp = {'sample_time': datetime.time(11, 55),
+               'barcode': '000028432',
+               'sample_date': datetime.date(2015, 6, 7),
+               'site_sampled': 'Stool'}
+        # transform results
+        res = [dict((k, r[0][k]) for k in exp.keys()) for r in res if r != []]
```
mortonjt: Haven't tested, but this could also be rewritten as
```
res = [{k:r[0][k] for k in exp.keys()) for r in res if r != []}]
```

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -312,139 +318,137 @@ def test_getAnimalParticipantsNotPresent(self):
...
+        # only look at those fields, that are not subject to scrubbing
+        self.assertIn(exp, res)
 
         i = "d8592c74-9694-2135-e040-8a80115d6401"
-        res = self.ag_data.getParticipantSamples(i, "REMOVED-0")
+        # collect all results for ALL names for this ag_login_id and remove
+        # field "notes" since it gets scrubbed.
+        names = list(set([name for name in
```
josenavas: I don't think you need the call to list, since you can iterate over the elements of a set too. I would rewrite this as:

```python
names = set(ag_data.get_participant_names_from_ag_login_id(i))
```

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -460,7 +464,9 @@ def test_getAvailableBarcodes(self):
 
         i = "d6b0f287-b9d9-40d4-82fd-a8fd3db6c476"
         res = self.ag_data.getAvailableBarcodes(i)
-        exp = ['000028434']
+        exp = [x['barcode'] for x in
```
josenavas: Are the barcodes being randomized for a given ag_login_id?

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -610,70 +668,92 @@ def test_checkPrintResults_invalid_ids(self):
...
-               'd8592c74-84a5-2135-e040-8a80115d6401', 'city': 'REMOVED',
-               'country': 'REMOVED', 'email': 'REMOVED', 'name': 'REMOVED',
-               'state': 'REMOVED', 'zip': 'REMOVED'}
-        self.assertEqual(exp, obs)
+        ag_login_id = 'd8592c74-84a5-2135-e040-8a80115d6401'
+        obs = self.ag_data.get_user_info(
+            self.ag_data.get_supplied_kit_id(ag_login_id))
+        # unfortunatly, most fields are scrubbed in the database, thus we
+        # cannot compare them over DB versions
+        exp = {'ag_login_id': ag_login_id,
```
josenavas: Can we check in this case that at least all the keys that we are expecting are there?
It will be something like:
```python
exp = ['address', 'ag_login_id', 'city', 'country', 'email', 'name', 'state', 'zip']
self.assertItemsEqual(obs, exp)
```
Although `obs` is a dictionary, when used in `assertItemsEqual` it only looks at the keys.

amgut/lib/data_access/test/test_survey.py
```diff
@@ -230,6 +229,56 @@ def test_american_name(self):
...
+        survey.
+        """
+
+        # Create random string to test update happens
+        c = ascii_letters + '1234567890'
+        notes_test = ''.join([choice(c) for i in range(40)])
```
josenavas: Can you add a non-ascii character? It may help to find issues in the future.

amgut/lib/data_access/test/test_survey.py
```diff
@@ -449,15 +461,19 @@ def test_store_survey_edit(self):
         self.assertEqual(obs, exp)
 
         obs = ag_data.getConsent(survey_id)
+        # change datatypes as postgres does:
```
josenavas: Shouldn't need to change datatypes as psycopg2 does that on the back.

amgut/lib/data_access/ag_data_access.py
josenavas: Agree that subclassing is outside the scope of this PR. However, I do think that prefixing the methods names with something like `utest_` ot `ut_` would be useful, so it will be really easy to see when a dev is using one of these methods outside of unit testing.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
...
+            TRN.add(sql, [ag_login_id])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('ag_login_id not in database: %s' %
+                                 ag_login_id)
+            return info[0][0]
+
+    def get_participant_names_from_ag_login_id(self, ag_login_id):
+        """ Returns all participant_name(s) for a given ag_login_id.
+        For unit testing only!
```
wasade: Some of these methods seem like they would either already exist, or at least have utility outside of unit tests?

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -90,49 +93,48 @@ def test_getAGBarcodeDetails(self):
...
             'other': None,
             'moldy': None,
             'refunded': None,
             'ag_kit_barcode_id': 'd8592c74-7e36-2135-e040-8a80115d6401',
             'date_of_last_email': None,
-            'other_text': 'REMOVED'
         }
-        self.assertEqual(obs, exp)
+        # only look at those fields, that are not subject to scrubbing
+        self.assertEqual(dict((k, obs[k]) for k in exp.keys()), exp)
```
wasade: fyi, same as: `self.assertEqual({k: obs[k] for k in exp}, exp)`

josenavas:
@sjanssen2 is there anything else that you need here or can you address the comments? Do you need help fixing any of the tests?

sjanssen2:
I am looking into the same things for labadmin right now ... but I can 
first do the review

josenavas:
Thanks @sjanssen2 let's get my other PR merged so I can generate another dump that I can send you, in which the test user for labadmin will be fixed - it may simplify some of your work. Do you agree?

amgut/lib/data_access/test/test_survey.py
sjanssen2: no. I specified types as follows:
`survey object, survey_id : str, notes_test : str and consent : dict of the newly created survey.`

amgut/lib/data_access/ag_data_access.py
sjanssen2: I changed the function name from "random" to "arbitrary" as suggested

amgut/lib/data_access/ag_data_access.py
sjanssen2: I prefixed those function names with "ut_"

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
...
+    # following are DB access functions only used for unit testing:
+
+    def get_random_supplied_kit_id_scanned_unconsented(self):
+        """ Returns randomly chosen supplied_kit_id and barcode which has been
+        scanned but is without consent.
+        For unit testing only!
+
+        Returns
+        -------
+        str: [supplied_kit_id, barcode]
```
sjanssen2: good catch

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
...
+        Returns
+        -------
+        str: barcode
+            Example: '000032951'
+
+        Raises
+        ------
+        ValueError
+            If no barcodes can be found in the DB."""
+        depos = 'TRUE'
```
sjanssen2: cool. I didn't know this and wanted to be on the save side. Removed those three lines.

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: no, I think they are stable, but their availability changes

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: good idea.
I don't like this implicitness, thus I explicitly compare against obs.keys() That will make it easier to read the code

amgut/lib/data_access/test/test_survey.py
sjanssen2: removed it

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
             sql = "SELECT deposited FROM ag.ag_kit_barcodes WHERE barcode = %s"
             TRN.add(sql, [barcode])
             return TRN.execute_fetchlast()
+
+    # following are DB access functions only used for unit testing:
+
+    def get_random_supplied_kit_id_scanned_unconsented(self):
+        """ Returns randomly chosen supplied_kit_id and barcode which has been
```
sjanssen2: no it is not. Depends on internal DB ordering. I changed the name to "arbitrary"

amgut/lib/data_access/ag_data_access.py
```diff
@@ -969,3 +969,303 @@ def is_deposited_ebi(self, barcode):
...
+                     AND ag.ag_kit_barcodes.survey_id IS NULL
+                     LIMIT 1"""
+            TRN.add(sql, [])
+            info = TRN.execute_fetchindex()
+            if not info:
+                raise ValueError('No kits found.')
+            return info[0]
+
+    def get_random_handout_printed_min6_supplied_kit_id(self):
+        """ Returns a random supplied_kit_id with printed results
```
sjanssen2: see above

amgut/lib/data_access/ag_data_access.py
sjanssen2: I checked with existing methods, there should be no overlap. The cleanest solution would be to move those functions into their own file but this is out of scope for this PR.

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: cool. I changes this at several locations

amgut/lib/data_access/test/test_survey.py
sjanssen2: and put it back in, since the test fails otherwise

amgut/lib/data_access/test/test_survey.py
sjanssen2: no I can't, because the code cannot cope with non ascii chars right now. I open issue #646

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: you are right

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: personally, I think my version is more readable. Thus, if you don't insist, I will leave it as is.

sjanssen2:
@josenavas @wasade @mortonjt I think this is ready to go, but please double check if I address all you comments appropriately

amgut/lib/data_access/test/test_ag_data_access.py
```diff
@@ -90,49 +93,48 @@ def test_getAGBarcodeDetails(self):
...
         exp = {'ag_kit_id': 'd8592c74-84bb-2135-e040-8a80115d6401',
-               'supplied_kit_id': 'tst_ODmhG',
-               'swabs_per_kit': 1,
+               'supplied_kit_id':
+               self.ag_data.ut_get_supplied_kit_id(
+                'd8592c74-84ba-2135-e040-8a80115d6401'),
+               'swabs_per_kit': 1L,
                'verification_email_sent': 'n',
-               'kit_verification_code': 'f4UjhV4B',
-               'kit_password': '$2a$12$LiakUCHOpAMvEp9Wxehw5OIlD/TIIP0Bs3blw18'
```
josenavas: Kit password should be the same for all kits ('test') so this value should be able to be tested. Commenting only once, but change in all that apply.

amgut/lib/data_access/test/test_ag_data_access.py
josenavas: Does it change based on the order in which the tests are executed?

The problem that I'm seeing here is that at the end the function self.ag_data.getAvailableBarcodes is going to be similar to self.ag_data.ut_get_barcode_from_ag_login_id which at the end is not executing any unit testing correctly, but rather checking that the "same" function returns the same value twice

amgut/lib/data_access/test/test_survey.py
josenavas: Thanks for opening the issue - can you also add a TODO in this code to make sure that when we solve #646 we add a non-ascii character in here?

amgut/lib/data_access/test/test_survey.py
josenavas: That's really weird - ok then - thanks!

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: I hope it is not dependent on execution order, but given the current quality and absence of unit tests I cannot exclude it. I'd prefer to leave all tests is for this PR but encourage everyone to contribute to more unit tests for American Gut

amgut/lib/data_access/test/test_survey.py
sjanssen2: done

mortonjt:
I'm ok with the PR as is

amgut/lib/data_access/test/test_ag_data_access.py
sjanssen2: agreed. I put password back in. And it should be the only place where we compare password string for unit tests

PR #647, opened 2017-03-28
enforcing usage of tornado 4.4.2 instead of 3.2.2

sjanssen2:


josenavas:
@wasade do you see any issue with this? if not I would be ok with merging it.

PR #648, opened 2017-03-28
Lowering the emails

josenavas:
There was an assumption on the labadmin code (and it is being ensured) that the emails are lowercase. However, that was not being enforced on the scrub script, which made fixing the tests quite interesting, since this was a bit hidden.


sjanssen2:
Since email addresses are not case sensitive, it should be totally fine. OK with me. And since this is a very small change it is ready to merge!

wasade:
google [suggests](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=email+case+sensitive&*) emails are case sensitive but that the domain is case insensitive?

sjanssen2:
I'd read this article as there is no practical negative impact if we restrict ourselves to limit email address to all lower cases.

wasade:
okay

PR #650, opened 2017-03-30
Update plate mapper branch to master

josenavas:
I'm just updating the plate mapper branch to the latest master - I didn't change any code, just moved the patch 40 from the plate mapper branch to be named 41 since there is already a patch 40 on the master branch to avoid merge conflicts.


josenavas:
@sjanssen2 @antgonza @wasade any of you available to merge?

PR #653, opened 2017-04-06
Sample plate fixes

josenavas:
A couple of changes in the patch.


antgonza:
👍 , do we need to reviews?

josenavas:
I don't think so - there are not that of a big changes

Issue #196, opened 2014-10-04
Remove 'Unspecified' from multiple select check boxes


wasade:
as title says

PR #654, opened 2017-04-07
WIP: Geocoder

sjanssen2:
See issue #196 in labadmin.
I want to move geocoder code into amgut and import it from labadmin. I intend to replace home brewed code with external libraries.


sjanssen2:
Has anyone an idea why geocoder tests fail on Travis but pass on my machine? Could it be that API calls from within Travis are blocked?

PR #657, opened 2017-04-11
Separating target gene and subfragment

josenavas:
I need to do this small change in the DB to support one of the comments from @wasade - should be fast


josenavas:
Thanks!!

PR #659, opened 2017-04-11
Adding qPCR values

josenavas:
Forgot to add a couple of columns

### DIFF
--- a/amgut/db/patches/0041.sql
+++ b/amgut/db/patches/0041.sql
@@ -479,6 +479,8 @@ CREATE TABLE pm.shotgun_library_prep_kit (
      water_volume_nl              real  NOT NULL,
      shotgun_i5_index_id          bigint  ,
      shotgun_i7_index_id          bigint  ,
+     qpcr_concentration           real ,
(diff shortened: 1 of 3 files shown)


amgut/db/patches/0041.sql
```diff
@@ -479,6 +479,8 @@ CREATE TABLE pm.shotgun_library_prep_kit (
      water_volume_nl              real  NOT NULL,
      shotgun_i5_index_id          bigint  ,
      shotgun_i7_index_id          bigint  ,
+     qpcr_concentration           real ,
```
antgonza: just want to confirm that real is fine and not double?

josenavas: I would say so - we've been using real for everything. Any concern?

antgonza: just making sure ...

wasade: real looks like it's comparable to single precision float but I didn't check if its actually ieee 754 compliant or whatever that spec is

PR #663, opened 2017-04-14
Fixing run structure

antgonza:


josenavas:
This have been done between 3 devs that we have been reviewing at the same time - hence merging.

PR #668, opened 2017-04-21
Adding good default values

josenavas:


amgut/db/patches/0041.sql
```diff
@@ -841,18 +844,19 @@ INSERT INTO pm.sample (sample_id, is_blank, details)
 
 -- Add a plate
 INSERT INTO pm.plate_reader (name, notes)
-    VALUES ('PR1234', 'Standard plate reader');
+    VALUES ('Synergy HT', 'Knight Lab Plate reader');
 
 -- Add a mosquito machine
 INSERT INTO pm.mosquito (name, notes)
-    VALUES ('Mosquito1', 'Standard mosquito machine');
+    VALUES ('Mosquito left', 'Mosquito machine on the left'),
```
wasade: These are the actual names from the wetlab? They'll need to retain those names...

josenavas: yeah - that is the current names that they use for the mosquito...

Issue #669, opened 2017-05-05
clean DB


sjanssen2:
Before we add in the new relation "source" we need to clean our DB to make it easier to understand for us and additional new devs.
I spent the last two days sifting through our codebase (master branches) of american-gut-web and labadmin and understand which tables are no longer used. I am quite confident about my findings, but I definitively need a two persons review before we actually drop the selected tables.

The following tables are not mentioned (by grep'ing case insensitive their raw name):
	ag_animal_survey
	ag_human_survey
	ag_import_stats_tmp
	ag_map_markers
	ag_participant_exceptions
	ag_survey_answer
	ag_survey_multiples
	ag_survey_multiples_backup
	barcode_exceptions
	controlled_vocabs
	controlled_vocab_values
	survey_response_types

I cannot see why those two tables are needed, because...
	duplicate_consents:
		labadmin/knimin/lib/data_access.py: format_survey_data(): line 726:
			here is the only place that we READ from this table. If there are no WRITES, why do we need this static information?

	promoted_survey_ids:
		american-gut-web/amgut/lib/ag_data_access.py: deleteAGParticipantSurvey(): line 325
			here is the only place where we DELETE from this table, when a survey (and I figure the consent) is deleted. Can't we simply delete the full table, i.e. for ALL users?

Please @josenavas @wasade review soon, since this is really time critical to get the metadata pulldown right.

sjanssen2:
here is how I got the number of occurrences of a table name in our codebases:
# get all tables:
psql --dbname=ag_test --password --host=localhost --username=sjanssen -c "\dt *." | grep -v "^ pg_catalog" | grep -v "information_schema" | grep -v "^ public" | grep "| table |" | cut -d "|" -f 1,2 | tr -d " " | tr "|" "." > ~/Desktop/tables.txt

# check occurence of table name:
grep "ag_map_markers" `find . -name "*.py" | xargs` -c | cut -d ":" -f 2 | sum.pl

# list all code files:
files=`find . -name "*.py" | xargs`; 

# get counts about table name occurence in code files:
files=`find . -name "*.py" | xargs`; for tabname in `cat ~/Desktop/tables.txt | cut -d "." -f 2`; do res=`grep "$tabname" $files -c -i | cut -d ":" -f 2 | sum.pl`; echo "$res $tabname"; done | cut -d ":" -f 2 | sort -n


# sum.pl
Is a small Perl script:
```
#!/usr/bin/env perl

use strict;
use warnings;
[...]

PR #670, opened 2017-05-05
Clean db

sjanssen2:
Remove unused tables, see issue #669


josenavas:
@sjanssen2 please read the "Making Database Changes" section of [Qiita's contributing file](https://github.com/biocore/qiita/blob/master/CONTRIBUTING.md). That applies to all our projects that involve DB but it is on the Qiita as we developed those docs for that project and we didn't want to duplicate the contents of that file on multiple repos.

Basically, you need to update the dbschema files (dbs and html). That is how we avoid them going out of sync, and that's why I pointed out on my email that the first task to perform is sync the dbschema files with the DB. The more they go out of sync the harder will be to sync them again.

amgut/db/patches/0041.sql
```diff
@@ -0,0 +1,24 @@
...
+DROP TABLE barcodes.barcode_exceptions;
+DROP TABLE ag.controlled_vocab_values;
+DROP TABLE ag.controlled_vocabs;
+
+-- we only delete information from this table, but never read or write. Can't we simply delete the full table then?
+DROP TABLE ag.promoted_survey_ids;
+
+-- a table we only read from, but never write. Thus, it is static. Can't we correct information elsewhere and get rid of this table?
+-- see discussion: https://github.com/biocore/american-gut-web/issues/669
+--DROP TABLE ag.duplicate_consents;
```
josenavas: Remove commented code

wasade: :+1: and or remove that table?

wasade:
Seems good barring @josenavas's comments

sjanssen2:
I am running out of battery power :-(

amgut/db/patches/0041.sql
sjanssen2: I am back to power now.
I think that this table will contradict with what we are going to implement with the many-to-many relation "source". And since @josenavas thinks that those information have never been validated, I'd like to remove this table. However, @EmbrietteH and @wasade needs to confirm this please.

wasade:
Okay that's fine

sjanssen2:
I cannot find version 6.1.0 of DbSchema as a downloadable package, thus I am not able to create the HTML. Can someone else step in or point me to a download location please.

sjanssen2:
I manually compared ag.dbs and actual table columns (with pgAdmin) and modified our schema where it was not in sync with the DB. Should be fine now.

sjanssen2:
Hi @wasade . Thanks for the late merge!
Do you know what was wrong with our duct tape solution, other than it was duct tape? Did it return correct results? Looks like I wasn't in the full loop of critique about this piece of code.

wasade:
I never saw results it produced from the live db. @EmbrietteH, did you get a chance to review records from the "ducttape" for pulling metadata from participants with multiple distinct surveys (eg main, fermented food, etc)?

I'm not sure what you mean by wrong with the solution? It was a hack to mimic the many-many relationships iirc

sjanssen2:
I want to double check that the SQL code you created for this solution returns correct data.
Before I left to Germany, I sent an example pulldown to @EmbrietteH but never got a response.

EmbrietteH:
That is incorrect. I did look. The first one produced incorrect pull-down
results; I talked to you about it, you fixed it, and the second one looked
sane, and I confirmed. Unless there is something else I am not
remembering...

sjanssen2:
Sorry, I was looking in my mail box for statements why my pulldown didn't work. You did reply me - my fault.

EmbrietteH:
No worries. I usually prefer to just talk face to face as its quicker, but
you're right then you don't have then paper trail. IIRC, some the barcodes
didn't pull down so there were unexpected failures. I don't recall what the
exact problem you IDd was...

PR #672, opened 2017-05-08
WIP: Db source

sjanssen2:
adding a new many-to-many relation for sources into the DB.
~~Work in progress: do not merge!~~


sjanssen2:
Although this is work in progress, your early feedback is very welcome @josenavas @wasade

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
+CREATE TABLE ag."source" (
+	source_id            uuid DEFAULT ag.uuid_generate_v4() ,
+	source_name          varchar(100)  ,
```
josenavas: Is the 100 chars limitation implemented in the GUI? If not either implement there or remove here.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
+CREATE TABLE ag."source" (
+	source_id            uuid DEFAULT ag.uuid_generate_v4() ,
```
josenavas: `uuid_generate_v4` is a function in the `ag` schema? PostgreSQL is capable of generating those UUID.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
...
+	source_name          varchar(100)  ,
+	ag_login_id          uuid  ,
```
josenavas: This should be a FK.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
+CREATE TABLE ag."source" (
```
josenavas: Double quotes are not needed (applies to all occurrences of table names).

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
...
+ );
+
+CREATE INDEX idx_source_1 ON ag."source" ( ag_login_id );
+
+COMMENT ON TABLE ag."source" IS 'A "source" is the human, animal or environment that was swapped for microbial analysis.';
+COMMENT ON COLUMN ag."source".source_id IS 'A unique ID to identify the source.';
+COMMENT ON COLUMN ag."source".source_name IS 'Sources can come with names, e.g. names for the humans that have been swapped.';
+COMMENT ON COLUMN ag."source".ag_login_id IS 'Points to the "user" that owns this source.';
+
+CREATE TABLE ag.source_barcodes_surveys (
```
josenavas: I think all columns of this table should be NOT NULL.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
...
+	source_id            uuid  NOT NULL,
+	barcode              varchar  ,
+	survey_id            varchar
```
josenavas: This column should be a FK.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
...
+       refunded,
+       deposited,
+			 ag_login_id,
+			 participant_name
+FROM participant_barcodes
+JOIN multiple_ids USING (ag_login_id, participant_name)
+FULL JOIN ag.ag_kit_barcodes USING (barcode);
+
+-- enrich with ag_login_id and participant_name; the later becomes source_name
+CREATE OR REPLACE TEMP VIEW foo AS
```
josenavas: Please provide useful names to the tables (even if they're temp) so a reader can understand better the code.

amgut/db/0042.sql
```diff
@@ -0,0 +1,129 @@
...
+)
+INSERT INTO ag.source_barcodes_surveys (source_id, survey_id)
+SELECT source_id, survey_id FROM ins1 JOIN ag.ag_login_surveys ON (ins1.ag_login_id = ag_login_surveys.ag_login_id AND ins1.source_name = ag_login_surveys.participant_name);
+
+-- fill many-to-many relation source_barcodes_surveys
+INSERT INTO ag.source_barcodes_surveys (source_id, barcode, survey_id)
+SELECT source_id, barcode, survey_id
+FROM foo
+JOIN ag.source ON (foo.ag_login_id = ag.source.ag_login_id AND
+									 foo.participant_name = ag.source.source_name);
```
josenavas: Fix indentation.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -323,7 +340,10 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
                 self.deleteSample(info['barcode'], ag_login_id)
 
             # Delete last due to foreign keys
-            sql = "DELETE FROM ag_kit_barcodes WHERE survey_id IN %s"
+            sql = "DELETE FROM ag.ag_kit_barcodes WHERE barcode IN %s"
```
josenavas: I think these 2 sql commands need to be in reverse order.

amgut/lib/data_access/ag_data_access.py
```diff
@@ -389,25 +409,30 @@ def logParticipantSample(self, ag_login_id, barcode, sample_site,
                          WHERE ag_login_id = %s AND participant_name = %s"""
 
                 TRN.add(sql, (ag_login_id, participant_name))
-                survey_id = TRN.execute_fetchindex()
-                if not survey_id:
-                    raise ValueError("No survey ID for ag_login_id %s and "
+                survey_ids = [x[0] for x in TRN.execute_fetchindex()]
```
josenavas: I think in the case that there are no survey ids this will fail. I think the correct thing to do is:
```python
survey_ids = TRN.execute_fetchindex()
if not survey_ids:
    raise...
survey_ids = [x[0] for x in survey_ids]
```

amgut/test/test_dbpatch_source.py
```diff
@@ -0,0 +1,104 @@
+from unittest import TestCase, main
```
josenavas: I don't think this tests need to be added to the repo.

amgut/db/0042.sql
sjanssen2: no longer used

amgut/db/0042.sql
sjanssen2: autogenerated by DbSchema, no longer used

amgut/db/0042.sql
sjanssen2: isn't that achieved by the later 
`ALTER TABLE ag.source_barcodes_surveys ADD CONSTRAINT fk_source_barcode FOREIGN KEY ( barcode ) REFERENCES barcodes.barcode( barcode );` like line?

amgut/db/0042.sql
sjanssen2: how would I do that?

amgut/db/0042.sql
sjanssen2: see above

amgut/db/0042.sql
sjanssen2: sure

amgut/db/0042.sql
sjanssen2: outdated

amgut/test/test_dbpatch_source.py
sjanssen2: I'd like to keep it in, because it should that no all tests did correctly revert DB changes.

amgut/lib/data_access/ag_data_access.py
sjanssen2: I changed as you suggested

amgut/lib/data_access/ag_data_access.py
sjanssen2: I switched order, however I cannot see any effect, which is good :-)

sjanssen2:
I think I reached a reasonable version that should be good for review. Please give me your comments @josenavas @wasade

amgut/db/patches/0042.sql
```diff
@@ -0,0 +1,64 @@
+-- create the new many-to-many relation between barcodes and surveys
+CREATE TABLE ag.source_barcodes_surveys (
+	barcode              varchar  ,
+	survey_id            varchar
```
josenavas: I think this column should also be a FK. You are only making barcode a FK, but this one too. Also, both should be NOT NULL.

amgut/db/patches/0042.sql
```diff
@@ -0,0 +1,64 @@
...
+FROM multiple_ids
+JOIN multi_source_bc USING (ag_login_id, participant_name);
+-- (1823 rows)
+-- (288 rows) distinct sources
+-- (871 rows) distinct barcodes
+-- (599 rows) distinct surveys !?! 652 --> because some sources with multiple surveys have no barcodes assignes!
+
+
+-- insert barcode survey (many-to-many) into newly created table
+DELETE FROM ag.source_barcodes_surveys; -- clear table prior to insertion
```
josenavas: you just created this table it should be empty. No need for this.

amgut/db/patches/0042.sql
```diff
@@ -0,0 +1,64 @@
...
+INSERT INTO ag.source_barcodes_surveys (survey_id, barcode)
+SELECT survey_id, barcode FROM (SELECT * FROM multi_source_bc_survey UNION SELECT * FROM single_source_bc_survey) AS foo;
```
josenavas: What is the `as foo` doing here?

amgut/test/test_dbpatch_source.py
josenavas: You can't enforce test order so I don't thing this is actually testing anything specific?

amgut/db/0042.sql
josenavas: That is only creating a FK for the column barcode not for the column ag_login_id...

amgut/db/0042.sql
josenavas: adding NOT NULL as it is done for the other columns? Also in DB schema you can specify all this values. Actually they SHOULD be specified, otherwise the documentation and the code go out-of-sync. DBSchema can generate the patch for the DB structure automatically. If it is wrong, that means that the DBSchema representation is also wrong. Make sure to keep them in sync.

amgut/lib/data_access/ag_data_access.py
josenavas: Actually it is bad. The call should've failed because there should've been a Foreign Key on the barcode. If it didn't fail before means that either the test is wrong or the FK is missing, both cases are bad. Please double-check.

amgut/test/test_dbpatch_source.py
sjanssen2: It becomes an issue if you run nosetests on your machine twice without resetting the DB manually.

amgut/db/patches/0042.sql
sjanssen2: that was indeed unnecessary. I removed it to make the SQL statement more easy

amgut/db/patches/0042.sql
sjanssen2: you are right, but I used it while developing the patch and re-started it several times.
It is now removed

amgut/db/patches/0042.sql
sjanssen2: ups. I forgot to add the FK line :-/
I also added the NOT NULL constraints

amgut/lib/data_access/ag_data_access.py
sjanssen2: Seems like the foreign key does not restrict deletion. Could you please check if I created those FK correctly in the patch?

sjanssen2: hmmm ... the table source_barcode_surveys must be able to store multiple barcodes for the same survey_id. Does a FK restriction on DELETE work if the column is NOT unique?

sjanssen2:
@wasade @josenavas any chance for a timely review? It is blocking my changes for labadmin because it needs the correct DB for Travis.

amgut/db/patches/0042.sql
```diff
@@ -0,0 +1,64 @@
...
+
+-- remove views
+DROP VIEW multi_source_bc_survey;
+DROP VIEW multi_source_bc;
+DROP VIEW single_source_bc_survey;
+DROP VIEW multiple_ids;
+
+-- remove incomplete information from other table:
+-- save that for another PR
+-- ALTER TABLE ag.ag_kit_barcodes DROP COLUMN survey_id;
```
wasade: why save this for another pr?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -323,8 +340,11 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
                 self.deleteSample(info['barcode'], ag_login_id)
 
             # Delete last due to foreign keys
-            sql = "DELETE FROM ag_kit_barcodes WHERE survey_id IN %s"
+            sql = """DELETE FROM ag.source_barcodes_surveys
+                     WHERE survey_id IN %s"""
             TRN.add(sql, [tuple(survey_ids)])
+            sql = "DELETE FROM ag.ag_kit_barcodes WHERE barcode IN %s"
```
wasade: What happens in the situation where the user only wants to delete one of multiple surveys, will this still remove the barcode from the ag_kit_barcodes table?

amgut/test/test_dbpatch_source.py
wasade: how sensitive is this to the state of the test db, and will this be heavily impacted when we reapply the scrub method again in the future?

sjanssen2: I figure we need to manually change those numbers to a new scrubbed version, because our DB will grow -> i.e. very sensitive to DB updates

amgut/db/patches/0042.sql
sjanssen2: to not break labadmin which still operates on these information

josenavas: One way of avoiding this is doing a PR to an alternative branch, so your PR to labadmin can be pointed to the alternative branch, so neither of this changes are merged into master until we are sure that both projects are updated.

amgut/lib/data_access/ag_data_access.py
sjanssen2: oh. I thought the whole ag_login_id should be removed (in the doc '# Remove user from new schema'), but you might be right that we only want to remove the "source", i.e. the combination of ag_login_id and participant_name. 
It should be sufficient to just remove the relation in source_barcodes_surveys, right?

amgut/test/test_dbpatch_source.py
josenavas: We have been doing an effort of adapting the test to be independent of the current status of the DB. I have the feeling this goes backwards with that. Also, if you run the tests twice in your machine and these tests fails, wouldn't that indicate that the problem is that there are tests in the repo that are not properly cleaning the environment?

josenavas:
@sjanssen2 Although I agree that this would block merging your code from labadmin, that doesn't mean that this is blocking development, as you can develop in parallel in the other project having this branch locally on your machine.

amgut/test/test_dbpatch_source.py
sjanssen2: You are right in all regards: 1) it is a step backward AND 2) we do have these improperly operating tests. Therefore, I suggest to keep this test until we are sure that both amgut and labadmin cleans properly after every test.

amgut/db/patches/0042.sql
sjanssen2: I opt for whatever is easiest to accomplish, since Rob stressed the time is critical and I don't want to spend too much dev time for managing separate branches

sjanssen2:
True, but in my experience, my local repo does not fail as often as Travis especially with geocoder and time outs

amgut/db/patches/0042.sql
josenavas: In that case I opt to merge in master with the new version and temporarily breaking labadmin. Note that you are saying that giving that this is time critical you are going to solve the issue soon, which means that the master branch can be broken in labadmin for 1 or 2 days while you finish this task, since we are not going to deploy in the live systems until both systems are fixed.

josenavas:
geocoder and timeout failures are stochastic and out of our control - shouldn't be a blocker for development...

sjanssen2:
master branch of labadmin seems to be broken already or are you able to successfully rerun it on Travis?

josenavas:
I can't access to the travis build. In any case, you can always do the PR in labadmin fixing this, and whenever that one is ready to merge we can merge this and simply rerun the tests in your labadmin PR.

sjanssen2:
please have a look at my latest PR in labadmin!

amgut/db/patches/0042.sql
wasade: @josenavas, the pulldown happens from labadmin

@sjanssen2, can you delete this comment, issue a separate PR with the `DROP` as a subsequent patch, and note that it cannot be merged until addressed in labadmin?

amgut/lib/data_access/ag_data_access.py
wasade: when we originally put this method together, we did not have multiple surveys in this fashion, so i think this may be a case of an outdated comment :/ i think the most correct thing is to a) remove the specific relation in `source_barcodes_surveys` and b) if and only if the associated barcodes are no longer associated to any survey in `source_barcodes_surveys` to then remove the barcode entry from `ag_kit_barcodes`. Does that make sense?

amgut/test/test_dbpatch_source.py
wasade: agree with @sjanssen2 on this.

amgut/db/patches/0042.sql
sjanssen2: deleted those three lines and move to PR #673

amgut/lib/data_access/ag_data_access.py
sjanssen2: Make sense. I changed the SQL by adding the condition that ag_kit_barcodes only get deleted if there is only one associated source_id left.

wasade:
I'm 👍

PR #673, opened 2017-05-09
Drop survey

sjanssen2:
Warning: don't merge until #672 is in!
directly relates to PR #672 and will remove the one column that would otherwise store partially replicate data.


sjanssen2:
Technically outdated

PR #674, opened 2017-05-10
Fixtests

sjanssen2:
Labadmin tests fail, because DB is in a different state if you either run amgut nosetests or not. This little PR should fix the different DB states.


sjanssen2:
@josenavas @wasade could you please quickly review to make fixing labadmin tests easier for me

PR #675, opened 2017-05-16
new master user for labadmin

sjanssen2:
Running unit-tests within labadmin resets the privileges for the only existing labadmin user 'test' to a very restrictive setting. For efficient manual debugging, one needs to add a new user with full privileges with two SQL statements.
To avoid this tedious task when re-populating the test DB, this patch adds another user 'master' for labadmin and sets its privileges to highest number (7).
This will save some time during development and testing.


sjanssen2:
@wasade @josenavas could you please review this small PR. It is NOT time critical.

josenavas:
@sjanssen2 The user needs to be added to the test database. If creating a patch, that will add it to the production database, creating a serious security problem.

Please review the code that is creating the test environment and when the test database is populated, you can add this user there. In that case, the 'master' user will be only created in the test environment, but not on production systems.

sjanssen2:
Ups, that would be really stupid :-/
I moved the function from being a patch to a python function in env_management, which is then called by the ag script if invoked as 'ag make test'.

amgut/lib/data_access/env_management.py
```diff
@@ -137,6 +138,29 @@ def make_settings_table():
         TRN.add(sql, args)
 
 
+def add_labadmin_superuser():
+    """Adds a new labadmin user names 'master' to the database and grants
+       highest privileges to this user. Password is 'password'."""
+    os.environ['PGPASSWORD'] = AMGUT_CONFIG.password
```
josenavas: Can you use the SQLConnectionHandler object that we have in the repository? That would avoid having to modify an environmental variable, plus it also will simplify error handling.

You will need to do the import inside this function, rather than at the top of the file, but I think that would be acceptable given the advantages of doing the call using the internal connection handlers.

scripts/ag
```diff
@@ -63,6 +63,8 @@ def make(db, force):
         # Test database includes initialization and settings table already
         click.echo("Populating the test database")
         populate_test_db()
+        click.echo("Adding labadmin superuser 'master', pwd is 'password'")
```
josenavas: Remove these two lines from this file.

amgut/lib/data_access/env_management.py
```diff
@@ -137,6 +138,29 @@ def make_settings_table():
...
+    proc = Popen(command, stdin=PIPE, stdout=PIPE)
+    retcode = proc.wait()
+    os.environ['PGPASSWORD'] = ''
+    if retcode != 0:
+        raise RuntimeError(("Could not add labadmin superuser 'master' to "
+                            "database %s: retcode %d") %
+                           (AMGUT_CONFIG.database, retcode))
+
+
 def populate_test_db():
```
josenavas: Add the creation of the master user inside this function, since it is only relevant for the test_db.

sjanssen2:
I think I addressed you comments by using TRN and moving the code into the populate_test_db function

EmbrietteH:
@wasade can you check and merge if ok?

PR #682, opened 2017-09-07
Remove incorrect vioscreen responses

wasade:
The labadmin data access method `store_external_survey` stored incorrect results due to an invalid split on `,` without respect to quoted values. All previously stored vioscreen entries are invalid.


EmbrietteH:
@josenavas @sjanssen2 can I get a review?

sjanssen2:
If it is `a off-by-one shift in the question / answer assocations.`, shouldn't we be able to correct those associations?
However, considering the migration to Viocare's API interfacing, it will save a lot of time if we don't have to check if existing data are consistent with new to be pulled down ones.
Therefore, I vote for deleting whatever is in our DB!

wasade:
The shift is lossy, and it's possible there is a > 1 shift

PR #683, opened 2017-09-15
Interactivity Refresh

justingeier:


tkosciol:
I was able to download this PR and set up the AG webserver and use interactivity with example data.

adswafford:
What else is needed to help this go live? Testing with more sample data, user feedback?

wasade:
This PR hasn't been reviewed so at a minimum that still needs to be done. I was chatting with @ElDeveloper and @nreeve17 about this today actually.

tkosciol:
it also most likely needs some changes on the backend to serve the data in a correct format.
originally, we were thinking about rolling it out to a sub-group of participants for initial feedback.
and one thing that is still WIP is Emperor integration into the results. we need to write some wrappers to e.g. highlight the sample of the participant in the PCoA.

justingeier:
Hi from Chicago! Please let me know if there is anything I can do to help
with this.

Justin

*Justin Geier*
Northwestern University | Evanston, Illinois | Class of 2019
Neuroscience Major | Spanish Minor
760.420.6761 | justingeier2019@u.northwestern.edu

PR #690, opened 2018-03-07
updated consent documentation for new participant page

nreeve17:
This PR updates the consent documentation for the new participant page.

### DIFF
--- a/amgut/lib/locale_data/american_gut.py
+++ b/amgut/lib/locale_data/american_gut.py
@@ -614,7 +614,6 @@ The following tests are available:</p>
 
 <p>You agree to allow your child to participate.</p>''',
 
-    'CONSENT_YOUR_CHILD_UPDATE': '''<p style="text-align: center;font-weight: bold;"></p>''',
     'PARTICIPATION_AGREEMENT': '''<p style="text-align: center;">University of California, San Diego</p>
 <p style="text-align: center;">
 Parent Consent for Child to Act as a Research Subject
--- a/amgut/lib/locale_data/american_gut.py
+++ b/amgut/lib/locale_data/american_gut.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# coding: utf-8
 from __future__ import division
 from amgut.lib.config_manager import AMGUT_CONFIG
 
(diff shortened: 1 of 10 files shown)


amgut/lib/locale_data/american_gut.py
```diff
@@ -330,158 +331,218 @@
...
+</p>
 
-<p>Research records may be reviewed by the UCSD Institutional Review Board.</p>
 
-<p style='font-weight: bold;'>Who can you call if you have questions?</p>
-<p>Dr Rob Knight and/or Gail Ackermann has explained this study to you and answered your questions. If you have other questions or research-related problems, you may reach Dr Knight at 858-246-1184</p>
+<p style='font-weight: bold;'><i>Who can you call if you have questions?</i></p>
+<p>
+If you have questions or research-related problems, you may reach Rob Knight at 858-246-1184 or contact the American Gut Project at 858-534-8739.
```
wasade: Nico, would it be possible to explicitly verify these phone numbers are correct with Sarah and/or Michiko?

amgut/templates/new_participant.html
```diff
@@ -131,7 +131,7 @@ <h3>{% raw tl['SEL_AGE_RANGE'] %}</h3>
 
 <div id="0-6-div" style='text-align: left;padding: 10px'>
 <div class='scrolltext'>
-{% raw tl['CONSENT_YOUR_CHILD']%}
+{% raw tl['CONSENT_YOUR_CHILD_UPDATE']%}
```
wasade: Is the `CONSENT_YOUR_CHILD` key no longer relevant?

wasade:
Just two comments. Sorry for the delayed review. I think this is nearly there

amgut/lib/locale_data/american_gut.py
nreeve17: No problem. I will double check with them.

amgut/templates/new_participant.html
nreeve17: I will go back and check what I did here. I put the CONSENT_YOUR_CHILD_UPDATE key when I was figuring out how html worked and wanted to make sure I wasn't breaking things. I will change it so that only the CONSENT_YOUR_CHILD key is there, I didn't add _UPDATE to the other keys, I don't think it makes sense to only have changed that one.

wasade: okay thanks!

amgut/lib/locale_data/american_gut.py
wasade: thank you!

PR #696, opened 2018-06-25
Fix to deleting survey and small visual changes

AlexanderFengFeng:
Implemented fix to bug involving deleting a survey profile with no samples. Also very slight visual changes for consistency to completed survey and participant overview.

### DIFF
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -314,7 +314,7 @@ class AGDataAccess(object):
             # removed and is has still revoked consent
             sql = """SELECT ag_login_id FROM ag.consent_revoked"""
             TRN.add(sql)
-            revoked = [result[0] for result in TRN.execute()[0]]
+            revoked = {result[0] for result in TRN.execute()[0]}
 
             sql = """SELECT survey_id, participant_email
                      FROM ag_login_surveys
@@ -357,7 +357,7 @@ class AGDataAccess(object):
                 # participant_name combination
                 if len(survey_ids) == 1:
                     sql = """DELETE FROM ag.ag_kit_barcodes
-                        WHERE barcode IN %s"""
+                             WHERE barcode IN %s"""
                     TRN.add(sql, [tuple(barcodes)])
 
             sql = "DELETE FROM ag_login_surveys WHERE survey_id IN %s"
@@ -367,7 +367,7 @@ class AGDataAccess(object):
                      WHERE ag_login_id = %s AND participant_name = %s"""
             TRN.add(sql, [ag_login_id, participant_name])
 
-            # only removes user from ag.consent_revoked if already there
+            # only inserts to ag.consent_revoked if not already there
             if ag_login_id not in revoked:
                 sql = """INSERT INTO ag.consent_revoked
                          (ag_login_id, participant_name, participant_email)
--- a/amgut/lib/data_access/ag_data_access.py
+++ b/amgut/lib/data_access/ag_data_access.py
@@ -310,12 +310,6 @@ class AGDataAccess(object):
     def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
         # Remove user from new schema
         with TRN:
-            # checks if user has previously been
-            # removed and is has still revoked consent
-            sql = """SELECT ag_login_id FROM ag.consent_revoked"""
-            TRN.add(sql)
-            revoked = {result[0] for result in TRN.execute()[0]}
-
             sql = """SELECT survey_id, participant_email
                      FROM ag_login_surveys
                      JOIN ag_consent USING (ag_login_id, participant_name)
@@ -366,6 +360,12 @@ class AGDataAccess(object):
             sql = """DELETE FROM ag_consent
                      WHERE ag_login_id = %s AND participant_name = %s"""
             TRN.add(sql, [ag_login_id, participant_name])
+
+            # checks if user has previously been
+            # removed and is has still revoked consent
+            sql = """SELECT ag_login_id FROM ag.consent_revoked"""
+            TRN.add(sql)
+            revoked = {result[0] for result in TRN.execute_fetchindex()}
 
             # only inserts to ag.consent_revoked if not already there
             if ag_login_id not in revoked:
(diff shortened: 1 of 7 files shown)


wasade:
Looks like some pep8 [issues](https://travis-ci.org/biocore/american-gut-web/builds/396541124#L746-L756)

amgut/lib/data_access/ag_data_access.py
```diff
@@ -357,16 +367,18 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
                      WHERE ag_login_id = %s AND participant_name = %s"""
             TRN.add(sql, [ag_login_id, participant_name])
 
-            sql = """INSERT INTO ag.consent_revoked
-                     (ag_login_id,participant_name, participant_email)
-                     VALUES (%s, %s, %s)"""
-            sql_args = [[ag_login_id, participant_name, pemail]
-                        for pemail in participant_emails]
-            TRN.add(sql, sql_args, many=True)
+            # only removes user from ag.consent_revoked if already there
```
wasade: isn't this inserting, not removing?

amgut/lib/data_access/ag_data_access.py
```diff
@@ -357,16 +367,18 @@ def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
...
+            if ag_login_id not in revoked:
```
wasade: should this check be done before the delete on line 359? eg:

```python
if ag_login_id in revoked:
    return  # no need to do anything
```

amgut/lib/data_access/survey.py
```diff
@@ -353,6 +353,12 @@ def store_survey(self, consent_details, with_fk_inserts,
             of the data to insert
         """
         with TRN:
+            # checks if user has previously been
+            # removed and has still revoked consent
+            sql = """SELECT ag_login_id FROM ag.consent_revoked"""
+            TRN.add(sql)
+            revoked = [result[0] for result in TRN.execute()[0]]
```
wasade: can this be made a set comprehension since it is used as a lookup? eg

```python
revoked = {result[0] for result in TRN.execute()[0]}
```

amgut/lib/data_access/ag_data_access.py
```diff
@@ -310,11 +310,18 @@ def get_all_handout_kits(self):
     def deleteAGParticipantSurvey(self, ag_login_id, participant_name):
         # Remove user from new schema
         with TRN:
+            # checks if user has previously been
+            # removed and is has still revoked consent
+            sql = """SELECT ag_login_id FROM ag.consent_revoked"""
+            TRN.add(sql)
+            revoked = [result[0] for result in TRN.execute()[0]]
```
wasade: can this be made a set comprehension since it is used as a lookup? eg

```python
revoked = {result[0] for result in TRN.execute()[0]}
```

amgut/lib/data_access/survey.py
wasade: ...actually, this can be done directly against the database by doing something like:

```python
sql = """select exists (select 1 from ag.ag_consent where ag_login_id=%s)"""
TRN.add(sql, [consent_details['login_id']])
exists = TRN.execute()[0]  # I think this is the necessary indexing...
```

Example from the ag test database below:

```sql
ag_test=# select exists (select 1 from ag.ag_consent where ag_login_id='ffffffff-7cf9-2135-e040-8a80115d6401');
 exists
--------
 f
(1 row)

ag_test=# select exists (select 1 from ag.ag_consent where ag_login_id='d8592c74-7cf9-2135-e040-8a80115d6401');
 exists
--------
 t
(1 row)
```

amgut/lib/data_access/ag_data_access.py
AlexanderFengFeng: Got it

amgut/lib/data_access/ag_data_access.py
AlexanderFengFeng: I put it in after these two statements

 363             sql = "DELETE FROM ag_login_surveys WHERE survey_id IN %s"
 364             TRN.add(sql, [tuple(survey_ids)])
 365 
 366             sql = """DELETE FROM ag_consent
 367                      WHERE ag_login_id = %s AND participant_name = %s"""
 368             TRN.add(sql, [ag_login_id, participant_name])

because I assumed they might still need to be called even if the id was still in consent_revoked. I figured they would cause an error if the revoked check needed to be put ahead of them.

amgut/lib/data_access/survey.py
AlexanderFengFeng: Hmm, I'm not sure how this exactly is supposed to work, since I don't quite know how to grab the consent details from the database (I get thrown the error that global name consent_details does not exist).

amgut/lib/data_access/survey.py
```diff
@@ -388,12 +388,30 @@ def store_survey(self, consent_details, with_fk_inserts,
...
-                         consent_details['survey_id'],
-                         consent_details['participant_name']))
+                    TRN.add("""INSERT INTO ag_login_surveys
+                                   (ag_login_id, survey_id, participant_name)
+                             VALUES (%s, %s, %s)""",
+                            (consent_details['login_id'],
+                             consent_details['survey_id'],
+                             consent_details['participant_name']))
+
+# checks if user has previously been
```
wasade: This comment looks like it should be indented

amgut/templates/human_survey_completed.html
```diff
@@ -15,4 +15,6 @@ <h2>{% raw tl['COMPLETED_HEADER'] %}</h2>
     <input type="hidden" name="go_home" value="Done">
     <input type="submit" value="Back to portal"></td>
 </form>
+<br>
```
wasade: I don't think this is needed?

amgut/templates/participant_overview.html
```diff
@@ -92,4 +92,6 @@ <h2>{% raw tl['OVERVIEW_FOR_PARTICPANT'] %} {{participant_name}}</h2>
         </table>
     </div>
 </div>
+<br>
+<br/>
```
wasade: same here

PR #697, opened 2018-06-27
Register kit rerender with previous details on invalid id/pw

AlexanderFengFeng:
Now should rerender with an error message and previous statements when submitting all entries with an invalid id or pw instead of returning to home page.


wasade:
I'm :+1: on this after #696 gets merged

PR #701, opened 2018-07-03
Added extract-survey script to output survey info

AlexanderFengFeng:
Outputs survey info to tsv


AlexanderFengFeng:
Seems all my problems are "No module named pandas"

AlexanderFengFeng:
@wasade would you be able to help me with this build fail

wasade:
It looks like pandas is [missing](https://travis-ci.org/biocore/american-gut-web/builds/399810812#L566). What about adding pandas to `setup.py`?

AlexanderFengFeng:
hm @wasade would you have any tips on how I could debug this?

AlexanderFengFeng:
@wasade got it to work by setting psycopg2==2.7.4 instead of psycopg2<2.7.5

amgut/lib/data_access/ag_data_access.py
```diff
@@ -211,6 +213,55 @@ def getAGBarcodeDetails(self, barcode):
                 raise ValueError('Barcode does not exist in AG: %s' % barcode)
             return dict(row[0])
 
+    def getAGSurveyDetails(self, survey_id, language):
```
wasade: It would be fantastic to have some unit tests on this

amgut/lib/data_access/ag_data_access.py
```diff
@@ -211,6 +213,55 @@ def getAGBarcodeDetails(self, barcode):
...
+        survey_id : str
+            the id of the survey group
+        language : str
+            the language the survey is intended for
+
+        Returns
+        -------
+        DataFrame
+            pandas DataFrame of sorted survey details
+        """
```
wasade:
```python
known_survey_ids= self.getKnownSurveyIds()
if survey_id not in known_survey_ids:
    raise ValueError("Unknown survey id: %s" % str(survey_id))
```

wasade:
```python
def test_getKnownSurveyIds(self):
    exp = {1, 2, 3, 4}  # whatever survey IDs we know for sure exist
    obs = ag_data_access.getKnownSurveyIds()
    self.assertEqual(obs, exp)
```

AlexanderFengFeng:
@wasade I added two methods to the AGDataAccess class which return known survey ids and languages and made three unit tests corresponding to the getAGSurveyDetails method and the two newly added methods. Do you have any suggestions for other unit tests/methods I could add?

wasade:
Very nice!

PR #720, opened 2018-11-08
Open bill of rights links in new tab/window

fedarko:
When the bill of rights PDF opens in the same tab/window, the user has to navigate "back" to return to the new participant page. However, when the user returns to the page, the containers of the consent/assent forms are kept closed and the radio button for age selection is kept selected -- which means the user has to manually select another age and then select the correct age option in order to re-open the correct consent/assent forms.

This is a partial fix for this problem, since it should make the bill of rights PDF open in a new tab or window (preserving the context of the original page). However, I think it'd also be a good idea to fix the problem of the consent/assent containers automatically closing in the future, in case the user ends up navigating back to this page anyway (due to either clicking a link in the sidebar, or their browser not recognizing `target="_blank"` for the bill of rights PDF).

### DIFF
--- a/amgut/templates/new_participant.html
+++ b/amgut/templates/new_participant.html
@@ -137,7 +137,7 @@
 </div>
     <div class="" id="consent">
         <hr>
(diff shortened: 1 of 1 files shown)


wasade:
Awesome, thank you!

PR #723, opened 2019-01-10
2.7 to 3.5.2

zoechallacombe:
I fixed all the str/byte errors that I could find and that were obviously causing problems, but I may missed some. Main error during testing is in line 14 of amgut/handlers/base_handlers.py, in get_current_user:
    skid = self.get_secure_cookie("skid")
    AttributeError: 'TestAddSampleOverview' object has no attribute 'get_secure_cookie'

Testing to see if this is reproducible in travis or if anything blatantly wrong sticks out.


zoechallacombe:
fixed .travis.yml using "sudo mount -o remount,size=25% /var/ramfs", hopefully it will run now.

PR #725, opened 2019-03-15
Labadmin buffer

wasade:


wasade:
Basic tables to hold status about labadmin pushes to Qiita.

The actual status messages are controlled by the front end, stored in the database for persistence and to facilitate async determination of state.

